PR #146, opened 2014-12-08
Fixes 8609 - validate the custom certificates passed in arguments on the installer

iNecas:


iNecas:
So the question is: rewrite or not? The facts are there. I'm not going to argue anymore: resolving this issue is more important to me than the language it's written in (although you know my preference:)

ehelms:
My preference hasn't been swayed either. I guess we let @jlsherrill break the tie and blame him there after ;)

jlsherrill:
eh, at this point I'm okay with it.  I'd prefer ruby, but i'm okay with bash.  The script isn't *that* complex.

mccun934:
can we get a rebase and merge as the tie has been broken

iNecas:
Rebased

ehelms:
@iNecas Looks like you you hit an RPM build failure.

iNecas:
right, updated

ehelms:
ACK

PR #155, opened 2015-01-06
Fixes #8843: katello-installer-base missing modules and hooks

ehelms:


jlsherrill:
Looking at the filelists from the generated rpms, this looks good to me: http://koji.katello.org/koji/taskinfo?taskID=206138

APJ

PR #164, opened 2015-01-15
Fixes #8849 - installs foreman_gutterball

dustints:
* installs foreman_gutterball
* fixes devel installation with gutterball


komidore64:
this looks sane to me!

dustints:
removing submodules that got added accidentally.

Puppetfile.lock
```diff
@@ -178,16 +178,6 @@ GIT
       puppetlabs-stdlib (>= 2.2.1)
 
 GIT
-  remote: https://github.com/puppetlabs/puppetlabs-postgresql
```
ehelms: Why is this getting removed here? I must be missing something.

dustints: didn't notice that! will restore back.
was trying to get librarian-puppet update going must have removed that while experimenting.

dustints: addressed

dustints:
@ehelms addressed

ehelms:
ACK

PR #165, opened 2015-01-16
remove puppetlabs-apache pinning

stbenjam:
I suspect we'll encounter the increased segfaults again though.


stbenjam:
Yup, not going to happen :disappointed: http://projects.theforeman.org/issues/7064

PR #166, opened 2015-01-22
fixes #9075 - hooks_dir should be hook_dirs

stbenjam:
See: https://github.com/theforeman/kafo/commit/1185cf813a392c63029a4ddcd4beca28b9e88034.  

capsule-installer hooks are in a common directory `/usr/share/katello-installer` and aren't being loaded.  It was working before because Kafo also looks in the installer directory itself, but now these are separated.


dustints:
ack, tested and looks good to me

stbenjam:
Thanks @dustints!

PR #167, opened 2015-01-26
Ref #9055 - puppet-katello: make packages depended on configurable

dustints:


config/answers.sam-installer.yaml
```diff
@@ -3,7 +3,9 @@ certs:
   generate: true
   deploy: true
   group: foreman
-katello: true
+katello:
+  package_names:
+    - katello-sam
```
ehelms: I think you'd want to include also the `ruby193-rubygem-katello` package.

mccun934: katello-common also requires ruby193-rubygem-katello .. aren't you duplicating that dependency?

ehelms: Think of it less like dependencies, and more like just tracking and ensuring packages are installed to ensure a particular state of the application. We *don't* have to include both. We just do for Katello.

dustints:
@ehelms i think sam package already has rubygem-katello as an indirect dependency through foreman_sam.

sam -> foreman_sam ->  rubygem-katello

sam spec: https://github.com/mccun934/katello/blob/20150109-1447/deploy/katello.spec#L180
foreman_sam spec: https://github.com/Katello/foreman_sam/blob/master/rubygem-foreman_sam.spec#L29

config/answers.sam-installer.yaml
dustints: @ehelms might be tricky to require rubygems-katello because of scl_prefix here...have to think of a way to denote that in package_names

dustints: maybe a new param like scl_package_names

ehelms: Good point, that does complicate things a bit.

ehelms:
APJ - given the above

config/answers.sam-installer.yaml
dustints: @ehelms started on this https://github.com/Katello/puppet-katello/pull/57/files to add a param for packages that may need scl prefixing..

ehelms:
[test]

dustints:
@ehelms failing bats because consumer-cert will be obtained from capsule/proxy instead of
main katello server. https://github.com/Katello/puppet-katello/pull/53

ehelms:
@dustints yea, I think you'll need to pull in the puppet-certs and puppet-capsule changes with this as well

dustints:
@ehelms ah ic, updated with puppet-certs/puppet-capsule changes for deploying consumer cert on capsule

ehelms:
[test]

ehelms:
This should unblock these failing tests -- https://github.com/Katello/puppet-certs/pull/47 . Please test and ACK if it does.

dustints:
tested seems ok from install...lets see if bats concurs

stbenjam:
I'm curious how you created this PR - did you manually update each module?

I cannot get `librarian-puppet` to update anymore, because regardless of what module I want, it tries to bring theforeman-foreman up to 3.0.0 which we cannot do as it violates our lock on pl-apache 1.1.1.

stbenjam:
[test]

ehelms:
ACK for me

dustints:
@stbenjam yah manually...i've always had issues with librarian-puppet resolving dependencies.

is that fine that its a manual update?

stbenjam:
For this, ACK from me.

I'll need to make some changes to Puppetfile to get librarian-update to work again, but I think I know whats wrong now.

Maybe for 2.3 "technical debt" work we can address versioning modules and managing the deps in metadata.json correctly...

PR #171, opened 2015-02-03
refs #8999,#9204,#9163 - updating foreman, service_wait, puppet, certs modules

jlsherrill:


ehelms:
ACK - manually tested with removing discovery prior to

PR #175, opened 2015-02-12
fixes #9075 - hooks_dir should be hook_dirs

stbenjam:
Backport of https://github.com/Katello/katello-installer/pull/166.  capsule-installer is missing hooks currently because of this.

---

See: theforeman/kafo@1185cf8.

capsule-installer hooks are in a common directory /usr/share/katello-installer and aren't being loaded. It was working before because Kafo also looks in the installer directory itself, but now these are separated.


ehelms:
APJ

stbenjam:
@ehelms I believe Jenkins will fail against KATELLO-2.1

ehelms:
Welp, ACK then :)

PR #187, opened 2015-03-09
refs #8175, #7064 - isolate consumer actions, and use foreman_proxy fork

ehelms:
This commit pulls in the changes needed to isolate consumer actions, as
well as pinning theforeman-foreman_proxy to our fork, which will avoid
pulling in a new puppetlabs-apache.  This avoids #7064, the case where
the installer segfaults due to poor garbage collection in 1.8.7.


stbenjam:
ACK

PR #189, opened 2015-03-11
Fixes #9718: Allow devel installer to use --reset

ehelms:


ehelms:
Anyone able to give this some review love?

hooks/boot/01-helpers.rb
```diff
@@ -20,18 +20,30 @@ def execute(commands)
...
 
       !results.include? false
     end
+
+    def command_exists?(command)
+      `which #{command} > /dev/null 2>&1`
+      $?.success?
+    end
+
+    def service_exists?(service)
```
stbenjam: I think this is misnamed, it's really checking if a process is running.

hooks/pre/10-reset_feature.rb
```diff
@@ -1,32 +1,45 @@
...
-  $?.success?
+def service
+  Kafo::Helpers.command_exists?('service-wait') ? 'service-wait' : 'service'
 end
 
 def stop_services
   Kafo::KafoConfigure.logger.info 'Ensuring services httpd and foreman-tasks are stopped.'
-  Kafo::Helpers.execute('service httpd stop && service foreman-tasks stop')
+  Kafo::Helpers.execute("#{service} httpd stop")
+  Kafo::Helpers.execute("#{service} foreman-tasks stop") if Kafo::Helpers.service_exists?('foreman-tasks')
```
stbenjam: Can we just use the new katello-service tool? It'll stop whatever's available, `katello-service --exclude mongod,postgresql` I think would do what you want and all the service commands could be removed from the other sections.

hooks/boot/01-helpers.rb
```diff
@@ -20,18 +20,30 @@ def execute(commands)
...
+      `ps cax | grep #{service} > /dev/null 2>&1`
```
ekohl: Do you really trust ps and grep for this?

hooks/pre/10-reset_feature.rb
ehelms: I think I can go that route -- I'll have to do an update to the devel puppet module to ensure katello-service gets installed first.

beav:
@ehelms is this still needed?

ehelms:
Still relevant but largely outdated

PR #193, opened 2015-03-18
Fixes #9809: Use 'yum localinstall' to install sub-man as well.

ehelms:
On non-rhel systems, users would need to ensure subscription-manager
got installed and then run the bootstrap RPM command. This will
change it such that the bootstrap RPM and it's dependencies are
attempted to be installed more cleanly for users.

### DIFF
--- a/hooks/post/10-post_install.rb
+++ b/hooks/post/10-post_install.rb
@@ -50,7 +50,7 @@ MSG
      parameters, see <%= color("capsule-installer --help", :info) %> and
      documentation for more info on setting up additional services):
 
-  rpm -Uvh http://#{fqdn}/pub/katello-ca-consumer-latest.noarch.rpm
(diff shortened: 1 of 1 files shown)


hooks/post/10-post_install.rb
```diff
@@ -50,7 +50,7 @@ def proxy_available?
      parameters, see <%= color("capsule-installer --help", :info) %> and
      documentation for more info on setting up additional services):
 
-  rpm -Uvh http://#{fqdn}/pub/katello-ca-consumer-latest.noarch.rpm
+  yum localinstall http://#{fqdn}/pub/katello-ca-consumer-latest.noarch.rpm
```
stbenjam: Needs `-y` to be unattended

ehelms:
@stbenjam please to re-review

stbenjam:
ACK

PR #200, opened 2015-03-31
Fixes #9964 - katello-installer should use apipie:cache:index

bagasse:


ehelms:
Please see -- https://github.com/Katello/katello-installer#filing-and-fixing-issues

bagasse:
> We have puppet-foreman locked to a particular version

Sorry, I missed that point.

ehelms:
@bagasse we've had to lock those modules due to the fact that the latest versions of puppetlabs-apache end up causing our installer to segfault on 1.8.7

ehelms:
We've recently updated the modules and this is now handled, see https://github.com/Katello/katello-installer/blob/master/modules/foreman/manifests/database.pp#L39

PR #201, opened 2015-04-01
Fixes #9988 - upgrade runs gb migrations

dustints:
* runs gutterball-db migrate during upgrade
* updates puppet-gutterball module
* other updates via librarian-puppet


hooks/pre/30-upgrade.rb
```diff
@@ -2,6 +2,22 @@ def migrate_candlepin
...
+  password = Kafo::Helpers.read_cache_data('gutterball_db_password')
+  migration_command = <<-END_OF_COMMAND
+/usr/bin/liquibase --driver=org.postgresql.Driver \
+        --classpath=/usr/share/java/postgresql-jdbc.jar:/var/lib/#{tomcat_dir}/webapps/gutterball/WEB-INF/classes/ \
+        --changeLogFile=db/changelog/changelog.xml \
+        --url=jdbc:postgresql:gutterball \
+        --username=#{username}\
+        --password=#{password}\
+        update
+END_OF_COMMAND
```
ehelms: @dustints what do you think about turning this into a script? Since it is being used in 2 places (here and in the puppet module) if we deploy and manage it as a script then we reduce managing this in 2 places. Further, the script could read from gutterball.conf on the filesystem since we are already storing the username/password within and then we don't have to worry about reading the values or caching them.

dustints:
@ehelms updated to use script that is created from Katello/puppet-gutterball#9

ehelms:
@dustints I think there is more here than you meant for there to be

dustints:
@ehelms is there a trick to plucking out just the changes for puppet-gutterball using librarian-puppet?
librarian-puppet seems to pull in everything else when I just want to update one module.

Also don't want to manually exclude what librarian-puppet pulls in, in the case that causes librarian-puppet to stop working due to conflicts.

stbenjam:
`librarian-puppet update <module>`

But, dependency resolution probably forces a few additional modules to be updated, not much you can do about it.

ehelms:
There are 2 empty directories included `modules/openssl` and `modules/nsstools`

dustints:
@stbenjam gotcha...yah i guess i'll just let librarian-puppet do it's thing.
@ehelms updated & removed the extra directories.

dustints:
[test]

stbenjam:
It's probably failing because of https://github.com/Katello/katello/pull/5185

dustints:
[test]

ehelms:
[test]

ehelms:
ACK

PR #205, opened 2015-04-08
Fixes #10096: Add use of '--profile' and update modules.

ehelms:
Updates the Foreman modules to 3.X and also Apache module that
previously led to segfaulting on EL6. This segfaulting is alleviated
by using Puppet's --profile option.


ehelms:
Key files to look at: 

Puppetfile
bin/katello-installer

I tested this locally on our vagrant centos6 box and saw no spikes in Puppet memory (held constant at around 8%). Speed wise, the install without `--profile` took 45ish minutes, while with these updates it dropped to 32ish minutes (I think in part due to newer Foreman module calling apipie index instead of running the full generation). So I did not see an increase in memory or time to install.

Puppetfile
```diff
@@ -19,9 +19,9 @@ mod 'theforeman-git', :git => 'https://github.com/theforeman/puppet-git'
 mod 'theforeman-tftp', :git => 'https://github.com/theforeman/puppet-tftp'
 
 # Top-level modules - Pinning & Forking due to #7064 FIXME
```
stbenjam: You can remove this comment too

bin/katello-installer
```diff
@@ -5,6 +5,10 @@ require 'kafo'
 
 CONFIG_FILE = './config/katello-installer.yaml'
 
+# Add --profile option on EL6 due to segfaulting
+os = `uname -a`
```
stbenjam: It doesn't happen so much with RHEL/CentOS but if a user compiles a custom kernel, this check might not catch it.

What do you think about looking at /etc/redhat-release instead and make sure it =~ `/ 6\./` or similar?

stbenjam:
I think you can also say this fixes [#7064](http://projects.theforeman.org/issues/7064) too.

bin/katello-installer
ekohl: Or look at the ruby version?

stbenjam: That's probably better, good idea

ehelms:
Updated

stbenjam:
@ehelms What about the issue number? I was tracking the segfault under 7064 before.

stbenjam:
ACK :tada:

jlsherrill:
ACK from me

ehelms:
I hope this means we can delete fork(s) of Foreman modules. Cheers!

PR #207, opened 2015-04-15
Use MariaDB instead of MySQL on EL7.

schiermi:
Fixes http://projects.theforeman.org/issues/8002

### DIFF
--- a/modules/mysql/manifests/params.pp
+++ b/modules/mysql/manifests/params.pp
@@ -25,20 +25,27 @@ class mysql::params {
 
   case $::osfamily {
     'RedHat': {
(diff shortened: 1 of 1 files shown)


ehelms:
Hi @schiermi , unfortunately you are attempting to modify a module that we pull into the git repository from its source (see https://github.com/Katello/katello-installer/blob/master/Puppetfile). If you want to see this changed, you'll need to submit a PR to the upstream puppet module and then we can pull it down into our installer. See https://github.com/Katello/katello-installer#module-bugs for more information.

ehelms:
Also, Katello does not currently support mysql/mariadb -- see http://projects.theforeman.org/issues/8125

schiermi:
Thank you for your fast response and clarification.

PR #211, opened 2015-04-20
Refs #10210: Explicitly set CRL to false since empty value is not the same.

ehelms:
I previously tried to use an empty value to mean empty string but that isn't the same so explicitly setting this value to false.


jlsherrill:
ACK

PR #212, opened 2015-04-21
fixes #10219 - use su instead of sudo

stbenjam:
To avoid the no tty problem.


ehelms:
APJ

ehelms:
do we need this for 2.2?

stbenjam:
I've never seen it, I bet QE was doing this remotely with no TTY.  I don't think it's urgent.

ehelms:
@stbenjam thanks! ACK

PR #213, opened 2015-04-21
fixes #10221 - handle services and failures better during upgrades

stbenjam:
We only need mongo started; having candlepin running generates some errors in the logs (temporary).  Also, we should just abort if one of the upgrade steps fails instead of silently continuing.

### DIFF
--- a/hooks/boot/01-helpers.rb
+++ b/hooks/boot/01-helpers.rb
@@ -7,7 +7,8 @@ class Kafo::Helpers
     end
 
     def log_and_say(level, message)
-      say "<%= color('#{message}', :#{level.to_s}) %>"
+      style = level == :error ? 'bad' : level
+      say "<%= color('#{message}', :#{style}) %>"
       Kafo::KafoConfigure.logger.send(level, message)
     end
 
--- a/hooks/pre/30-upgrade.rb
+++ b/hooks/pre/30-upgrade.rb
@@ -1,3 +1,13 @@
+def stop_services
+  Kafo::Helpers.execute('katello-service stop --exclude mongod')
+end
+
+def start_mongo
+  status = Kafo::Helpers.execute('service-wait mongod status')
+  Kafo::Helpers.execute('service-wait mongod start') unless status
+  Kafo::Helpers.execute('service-wait mongod status')
+end
+
 def migrate_candlepin
   Kafo::Helpers.execute("/usr/share/candlepin/cpdb --update --password #{Kafo::Helpers.read_cache_data('candlepin_db_password')}")
 end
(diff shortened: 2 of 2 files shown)


stbenjam:
Also :stop_services will fail on the capsule, we need katello-service there first.

hooks/pre/30-upgrade.rb
```diff
@@ -23,14 +27,24 @@ def migrate_foreman
...
-  send(step) unless app_value(:noop)
+  Kafo::Helpers.log_and_say :info, "Upgrade Step: #{step}#{noop}..."
+  unless app_value(:noop)
+    status = send(step)
+    fail_and_exit "Upgrade step #{step} failed. Check logs for more information." unless status
+  end
+end
+
+def fail_and_exit(message)
+  Kafo::Helpers.log_and_say :warn, message
```
ehelms: If we are failing and exiting, that wouldn't be an error?

hooks/pre/30-upgrade.rb
```diff
@@ -1,3 +1,13 @@
+def stop_services
+  Kafo::Helpers.execute('katello-service stop')
+end
+
+def start_mongo
+  status = Kafo::Helpers.execute('service-wait mongod status')
+  Kafo::Helpers.execute('service-wait mongod start') unless status
```
ehelms: Can systemd still timeout on us here?

hooks/pre/30-upgrade.rb
stbenjam: Fixed

hooks/pre/30-upgrade.rb
stbenjam: I've not seen that problem myself, does service-wait just need a longer timeout or something? We can't continue with the pulp migration unless mongo is confirmed started.

hooks/boot/01-helpers.rb
```diff
@@ -7,7 +7,8 @@ def module_enabled?(kafo, name)
     end
 
     def log_and_say(level, message)
-      say "<%= color('#{message}', :#{level.to_s}) %>"
+      style = level == :error ? 'bad' : level
```
stbenjam: Logger + colorization match for "info" but for whatever reason the color of error is "bad".

hooks/pre/30-upgrade.rb
stbenjam: I've been looking at making katello-service a little more intelligent. maybe one of the things I can add is an "--exclude" option so we can skip mongo in the first step in stopping everything else.

ehelms: I have seen this on slower VMs for EL7 because its more of a systemd issue. When we start mongod with upstart, it will just return thinking it has started mongo successfully, and then we use service-wait to poll for a condition before we mark it as 'truly' started. In the systemd case, it actually waits for the service to be entirely started and has a default 90 second timeout (which is the part I have run into on slower machines, https://bugzilla.redhat.com/show_bug.cgi?id=1040573)

ehelms:
@stbenjam will this be updated based on the katello-service PR?

stbenjam:
@ehelms Yup, once https://github.com/Katello/katello/pull/5181 is merged, I'll refactor this to use it so we don't stop/start mongo.

hooks/pre/30-upgrade.rb
stbenjam: So I exclude mongod from the stop now, so in MOST cases it will already be running by the time we check here. If it's not, and we hit this timeout problem, the pulp_migrate will fail, and the installer will exit, so at least it's not silent.

ehelms:
ACK

PR #216, opened 2015-04-27
fixes #10282 - migrate 2.1 answer files to 2.2

stbenjam:


stbenjam:
@ehelms What do you think of this approach? I figure we need an "add" operation, and it's the simplest to implement, then we could later add deletes and renames.

Maybe "migrations" could eventually be a part of Kafo itself, certainly it's a generic enough problem.

config/answers.katello-installer.yaml
```diff
@@ -36,16 +35,4 @@ foreman::plugin::default_hostgroup: false
...
-# Issue #4244 - Commenting these out due to the seg fault
-# caused on Ruby 1.8.7, the Katello RPM ensures these
-# are installed instead
-#foreman::compute::ec2: true
-#foreman::compute::gce: true
-#foreman::compute::libvirt: true
-#foreman::compute::openstack: false
-#foreman::compute::ovirt: true
-#foreman::compute::rackspace: false
-#foreman::compute::vmware: true
```
ehelms: I wonder if the --profile option we added allows us to re-enable these.

bin/katello-installer
```diff
@@ -2,12 +2,29 @@
 
 require 'rubygems'
 require 'kafo'
+require 'yaml'
 
 CONFIG_FILE = './config/katello-installer.yaml'
+ANSWER_FILE = './config/answers.katello-installer.yaml'
```
ehelms: I suppose this shouldn't change, but it could be gotten from the config file instead so there is one source of truth (and not having to worry about doing another sed replacement).

ehelms:
Did you try the dynamic case of migrating the existing on disk answers file as compared to the new one being laid down? I suppose there is no reason a user should disable gutterball,  but this would override that if they had?

stbenjam:
> Did you try the dynamic case of migrating the existing on disk answers file as compared to the new one being laid down?

The answer file packaged with the RPM isn't complete, so I think we need to do the diff ourselves, but we could have some tool to help generate the migrations based on the diff.

When we get to the more complicated cases we're going to have to make decisions - if a parameter was removed - was it really removed or just renamed? etc.  That's not automatable.

I want migrations to be part of Kafo and that's how we solve this upgrading problem... so we could have some proper tooling around it.

> I suppose there is no reason a user should disable gutterball, but this would override that if they had?

It shouldn't, the migration will only run once (on the first time running katello-installer after upgrading). If they later disable something it'll stay that way

ehelms:
Overall approach looks sound then, just my comments and I'd like to test this once.

ekohl:
Something that just came to mind, but is most likely already an issue: if something is an extra service and later disabled, then puppet won't manage it but it will still run. Suppose elasticsearch is removed, then it will still be in a users runlevel.

stbenjam:
> Something that just came to mind, but is most likely already an issue: if something is an extra service and later disabled, then puppet won't manage it but it will still run. Suppose elasticsearch is removed, then it will still be in a users runlevel.

Just a reality of using Puppet, when we remove elasticsearch one day (or anything else), we'll have to keep a module around to clean up and remove it.

stbenjam:
@ekohl I filed http://projects.theforeman.org/issues/10291 to remind us for that case later

@ehelms updated

stbenjam:
I also filed a bug about the general migration idea here: http://projects.theforeman.org/issues/10292

stbenjam:
[test] 

Scratch build didn't seem like it built correctly

stbenjam:
RPM building is not getting the changes because the diff is failing to apply: http://ci.theforeman.org/job/katello_packaging_build_matrix_rpm_katello_installer/920/label=rpmbuild,os=rhel6/console

    ERROR: certificate common name “github.com” doesn’t match requested host name “patch-diff.githubusercontent.com”.

stbenjam:
[test] again

ehelms:
I am hitting:

```
/sbin/katello-installer:21:in `block in <main>': undefined method `[]' for #<Puppet::Transaction::Report:0x00000001687440> (NoMethodError)
        from /sbin/katello-installer:18:in `glob'
        from /sbin/katello-installer:18:in `<mai
```

ehelms:
Oh, cause the config file won't have migrate options in previous versions like 2.1

stbenjam:
@ehelms Oh, I see - fixed.  katello-installer.yaml shouldn't be marked noreplace.

ehelms:
I think that makes sense. What about the password stored in there?

stbenjam:
> I think that makes sense. What about the password stored in there?

Oh, that's unfortunate. What's that a password to?

ehelms:
I believe you can use it to encrypt passwords in the answers file. Not sure
we are doing that that I know of. Kafo docs would have more

stbenjam:
Oh, I see.  

If Kafo is adding the password after the fact, %config at all is going to prevent the file from getting overwritten.  I guess this is something Foreman just doesn't care about because they don't use kafo between versions (despite the docs saying otherwise...)  :-1: 

Another problem to solve...

ehelms:
This looks good to me and worked as I expected. I ran into 2 issues that this revealed.

The first appears to be with gutterball itself and migrating:

```
[ WARN 2015-04-28 19:31:38 main]  /Stage[main]/Gutterball::Database/Exec[migrate database]/returns: Liquibase update Failed: Validation Failed:
[ WARN 2015-04-28 19:31:38 main]  /Stage[main]/Gutterball::Database/Exec[migrate database]/returns:      1 change sets check sum
[ WARN 2015-04-28 19:31:38 main]  /Stage[main]/Gutterball::Database/Exec[migrate database]/returns:           db/changelog/2014-12-03-15-39-add-event-status.xml::20141203153942-1::dgoodwin is now: 7:5682e551dd6fb1c61326d03bc0dab472
[ERROR 2015-04-28 19:31:38 main]  /Stage[main]/Gutterball::Database/Exec[migrate database]: Failed to call refresh: gutterball-db migrate returned 255 instead of one of [0]
[ERROR 2015-04-28 19:31:38 main]  /Stage[main]/Gutterball::Database/Exec[migrate database]: gutterball-db migrate returned 255 instead of one of [0]
```

The other is based on an assumption we are making that `gutterball-db` exists for the upgrade step, however, this file isn't created until the installer is actually run. Chicken and egg problem.

ehelms:
I should note that I saw migration errors because the puppet module itself attempts the migration later due to a refresh from laying down the `gutterball-db` file for the first time.

ehelms:
ACK for me as we are working out the other 2 issues

stbenjam:
@ehelms Before I merge - should we also migrate the katello-installer.yaml file in the same way?

Or solve it some other way? We still need to get the gutterball mapping in there some how :-\

ehelms:
Blast... good catch. Yea, since its yaml and the functionality should be re-usable I think that makes the most sense.

stbenjam:
Updated.  Because config file won't have the "migrate_dir", I moved it back into katello-installer.

I also needed to implement some simple deep(er) hash merging to migrate the `:mapping` section.  I'd rather use the deep_merge from ActiveSupport, so maybe we want to pull that in eventually? But that's a big library to pull into the non-SCL environment for just this..

ehelms:
ACK

PR #219, opened 2015-05-04
Fixes #10350: Set qdrouterd certs permission to qdrouterd user

ehelms:


stbenjam:
What about https://github.com/Katello/puppet-qpid/blob/master/manifests/router/config.pp#L28?

stbenjam:
Also, this will fail if the new packages aren't already in our repos - the `qdrotuerd` user doesn't exist on the packages that are out there in the wild now.

ehelms:
@stbenjam I found no difference for the config file ownership testing this out locally.

ehelms:
I should note I was testing against packages that have this error, so changing the crt/key ownership fixed them for me.

stbenjam:
Ah ok, well APJ

PR #226, opened 2015-05-19
Fixes #10538 - Adding capsule-remove script

chris1984:
Added capsule remove script to /bin and modifed spec file to include

PR #227, opened 2015-05-19
Fixes #10538 - Adding capsule-remove script

chris1984:
Added capsule remove script and modified .spec file

### DIFF
--- a/bin/capsule-remove
+++ b/bin/capsule-remove
@@ -1,0 +1,126 @@
+#!/bin/bash
+
+echo ""
+echo "WARNING: This script will erase many packages and config files."
+echo "Important packages such as the following will be removed:"
+echo ""
+echo "  * httpd (apache)"
+echo "  * mongodb"
+echo "  * tomcat"
+echo "  * puppet"
+echo "  * ruby"
+echo "  * rubygems"
+echo "  * All Katello and Foreman Packages"
+echo ""
+echo "Once these packages and configuration files are removed there is no going back."
+echo "If you use this system for anything other than Katello and Foreman you probably"
+echo "do not want to execute this script."
+echo ""
+read -p "Read the source for a list of what is removed.  Are you sure(Y/N)? " -n 1 -r
+echo    # (optional) move to a new line
+if [[ ! $REPLY =~ ^[Yy]$ ]]
+then
+    echo "** cancelled **"
+    exit 1
+fi
+
+echo ""
+echo "ARE YOU SURE?: This script permanently deletes data and configuration."
+read -p "Read the source for a list of what is removed.  Type [remove] to continue? " -r
+echo    # (optional) move to a new line
+if [[ ! $REPLY = remove ]]
+then
+    echo "** cancelled **"
+    exit 1
+fi
+
+CONFIG_FILES=(
+    /etc/pulp/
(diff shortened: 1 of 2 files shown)


ehelms:
Now that you have the spec file squared away, can you update the script based on the comments from https://github.com/Katello/katello-installer/pull/226 ? Otherwise, not everything will be removed properly for the current codebase (where we split out the capsule-installer).

bin/capsule-remove
```diff
@@ -0,0 +1,122 @@
...
+    /etc/pulp/
+    /usr/share/katello-installer/
+    /var/lib/puppet/
+    /var/lib/mongodb/
+    /var/lib/pulp/
+    /etc/httpd/
+    /usr/share/pulp
+    /etc/foreman-proxy
+    /var/lib/foreman-proxy
+    /etc/katello-installer
```
ehelms: This needs to be `/etc/capsule-installer`

bin/capsule-remove
```diff
@@ -0,0 +1,122 @@
...
+    /etc/foreman-proxy/
+    /etc/puppet/environments
+    /etc/pki/katello-certs-tools
+)
+
+LOG_FILES=(
+    /var/log/katello-installer/
```
ehelms: This needs to be `/etc/capsule-installer`

bin/capsule-remove
```diff
@@ -0,0 +1,122 @@
...
+    /etc/pki/katello
+    /root/ssl-build
+)
+
+# stopping all katello/foreman related services
+
+katello-service stop 
+
+echo "Removing RPMs"
+yum erase -y ${RPMS[*]}
```
ekohl: Won't this may remove a lot more than the user expected? If anything depends on python-requests for example.

bin/capsule-remove
```diff
@@ -0,0 +1,122 @@
...
+    `rpm -qa | grep ^python-webpy`
+    `rpm -qa | grep ^python-nectar`
+    `rpm -qa | grep ^python-saslwrapper`
+    `rpm -qa | grep ^python-amqp`
+    `rpm -qa | grep ^python-billiard`
+    `rpm -qa | grep ^python-semantic-version`
+    `rpm -qa | grep ^python-requests`
+    `rpm -qa | grep ^python-isodate`
+    `rpm -qa | grep gutterball`
+    `rpm -qa | grep $HOSTNAME`
```
ekohl: It is possible that a user has a cluster where the hostnames are simply ```1``` or ```01``` and then this can match a lot more than you expect (kernel with revision 201 for example). I'd recommend looking at the full hostname.

bin/capsule-remove
chris1984: imo this is a reasonable action, if the system is going to be used for anything else production wise then it would be safe to blow away the os or reinstall to make sure nothing is lingering.

bin/capsule-remove
chris1984: @ekohl updated the script to get hostname -f and assign it to FQDN and then use that when doing the rpm -qa to fulfill your request

ehelms:
This is looking good on my end, can you squash the commits into one?

ehelms:
Tested and worked well. ACK, merging thanks @chris1984 !

PR #230, opened 2015-06-02
fixes #10680 - use --profile for el6 devel install

jlsherrill:


ehelms:
APJ

ehelms:
This doesn't affect the job that is failing currently so I am good with it being merged (ACK)

PR #233, opened 2015-06-05
fixes #10716 - migrate pulp in post as well

stbenjam:
Pulp database is migrated before pulp-plugins-docker is installed during an uprade, which results in errors like this:

```
Importer docker_importer: no type definition found for docker_image
Jun 05 08:59:09 capsule.stbenjam.satellite.lab.eng.rdu2.redhat.com pulp15120: pulp.server.webservices.application:CRITICAL: *****************************************************
```

The pulp database should be migrated in post, as well, in case we installed any additional pulp plugins.

### DIFF
--- a/hooks/post/30-upgrade.rb
+++ b/hooks/post/30-upgrade.rb
@@ -1,9 +1,14 @@
 if app_value(:upgrade)
-  if Kafo::Helpers.module_enabled?(@kafo, 'katello')
-    noop = app_value(:noop) ? ' (noop)' : ''
+  if Kafo::Helpers.module_enabled?(@kafo, 'katello') || @kafo.param('capsule', 'pulp').value
+    Kafo::Helpers.log_and_say :info, "Upgrade Step: migrate_pulp...#{noop}"
+    Kafo::Helpers.execute('su - apache -s /bin/bash -c pulp-manage-db') unless app_value(:noop)
 
     Kafo::Helpers.log_and_say :info, "Upgrade Step: Restarting services...#{noop}"
     Kafo::Helpers.execute('katello-service restart') unless app_value(:noop)
+  end
+
+  if Kafo::Helpers.module_enabled?(@kafo, 'katello')
+    noop = app_value(:noop) ? ' (noop)' : ''
 
     Kafo::Helpers.log_and_say :info, "Upgrade Step: db:seed...#{noop}"
     Kafo::Helpers.execute('foreman-rake db:seed') unless app_value(:noop)


ehelms:
[test]

stbenjam:
[test]

ehelms:
Should we write up an issue for the Pulp puppet module to fire off a migrate if a new plugin is installed/enabled? Also, since the user would run `yum update` prior to running this upgrade, the ruby193-rubygem-katello package should install this plugin (https://github.com/Katello/katello/blob/master/rubygem-katello.spec#L61) and thus the initial migrate should cover this use case?

stbenjam:
Hmm, it doesn't seem to, it always needs 2 migrations for me. 

We could do it in Puppet, and probably makes sense to have a plugin architecture that handles this.  My main concern in doing something lik now is such a big change to the downstream installer, this PR is pretty low-impact.

ehelms:
I agree that doing it in Puppet should be the long term goal, just more curious why if the RPM is installing the plugin, the migration is not already handling this for us. These errors manifest post 2.2 upgrade once the user starts performing docker repository operations?

stbenjam:
I haven't tried upstream, all my focus has been on downstream, but it's there on every upgrade I tried, and the pulp-plugin-docker was already installed by the initial yum update.

Perhaps someone from pulp can clarify, but I always need two migrations.

I don't think we're guaranteed to always pull in the package by the yum update anyway, especially if plugins for pulp become optional, so I think it still makes sense to migrate in post.

ehelms:
On Fri, Jun 12, 2015 at 10:26 AM, Stephen Benjamin <notifications@github.com
> wrote:

> I haven't tried upstream, all my focus has been on downstream, but it's
> there on every upgrade I tried, and the pulp-plugin-docker was already
> installed by the initial yum update.
>

Adding code upstream for an issue we can't reproduce to try to fix
someone's downstream version of our code doesn't seem right.


> Perhaps someone from pulp can clarify, but I always need two migrations.
>
> I don't think we're guaranteed to always pull in the package by the yum
> update anyway, especially if plugins for pulp become optional, so I think
> it still makes sense to migrate in post.
>

Why wouldn't yum pull this in if it's a dependency? Theres a good chunk of
work to make plugins optional, and I haven't seen anything that puts that
on the horizon.

stbenjam:
> Adding code upstream for an issue we can't reproduce to try to fix
someone's downstream version of our code doesn't seem right.

I didn't say it wasn't upstream, I just haven't had any chance to try.

ehelms:
On Fri, Jun 12, 2015 at 11:07 AM, Stephen Benjamin <notifications@github.com
> wrote:

> Adding code upstream for an issue we can't reproduce to try to fix
> someone's downstream version of our code doesn't seem right.
>
> I didn't say it wasn't upstream, I just haven't had any chance to try.
>

Apologies, I meant this from a I haven't seen it in upgrade testing and no
users have reported it. However, if you only see that error when doing
docker operations then that could explain it. This would be a good
incentive to push to get https://github.com/Katello/katello-deploy/pull/128
in.


>  —
> Reply to this email directly or view it on GitHub
> <https://github.com/Katello/katello-installer/pull/233#issuecomment-111520737>
> .
>



-- 
Eric D. Helms
Red Hat Engineering
Ph.D. Student - North Carolina State University

stbenjam:
Since docker was introduced in 2.1, I don't know there were really many if any upstream users with a 2.0 capsule that upgraded to 2.1, and if they had problems they may have just run the installer again which would've made the problem go away.

We still have some time, I'll try to reproduce it on upstream. I asked on #pulp but no one answered.

@rbarlow have you seen this behavior where installing pulp-docker-plugins requires migrating twice?

rbarlow:
On 06/12/2015 11:25 AM, Stephen Benjamin wrote:
> @rbarlow <https://github.com/rbarlow> have you seen this behavior where
> installing pulp-docker-plugins requires migrating twice?

I have not seen issues requiring double-migration. Is this reproducible
with upstream Pulp?

-- 
Randy Barlow

stbenjam:
It's not, the root cause is actually that on the capsule, the plugin isn't installed until after the migration.  On Katello it's part of the `katello` RPM requirements so it gets pulled in during yum ugprade.

stbenjam:
[test]

ehelms:
[test]

ehelms:
ACK

PR #235, opened 2015-06-23
Module update.

ehelms:


stbenjam:
[test]

ehelms:
@stbenjam or @jlsherrill  are you OK with doing a mass module update like this?

stbenjam:
Fine for me, BATS is green. 

I was looking at foreman-installer the other day and I do wish we'd move to that model - not now, but in the medium-term (2.4?).

ehelms:
I would prefer that as well, however, I don't know what the scenario installer will mean so that is partially why I have avoided it.

PR #236, opened 2015-07-02
fixes #10986 - ensure noop is defined in upgrade hook

stbenjam:


hooks/post/30-upgrade.rb
```diff
@@ -1,4 +1,6 @@
 if app_value(:upgrade)
+  noop = app_value(:noop) ? ' (noop)' : ''
```
ehelms: How does moving it earlier help ensure its defined?

stbenjam: https://github.com/Katello/katello-installer/pull/236/files#diff-1453f17351c9fefb73f41a2dd0837b82R5 references it, and it was before it was defined.  Do you want a better commit message maybe?

ehelms: Ahh, yea thanks for pointing that out. I figured there was something I just wasn't seeing it.

ehelms:
APJ

stbenjam:
[test]

ehelms:
Since this is only used for upgrade, and we don't test upgrades through the CI system I am ok with this being merged.

stbenjam:
Ok, thanks @ehelms

PR #237, opened 2015-07-07
Fixes #11041: Setup GPG fragment properly for Apache 1.1.X module.

ehelms:
The Apache 1.1.X module does not contain the ability to concat fragments
directly onto the apache vhost which is what this version of the installer
contains.


ekohl:
Wouldn't it be possible to use a newer Apache module?

ehelms:
Note that this is going into our 2.2 branch for our 2.2.3 release so I am
trying to minimize the change.

jlsherrill:
ACK

ekohl:
@ehelms until you mentioned it on IRC I didn't notice it, so that makes sense.

PR #240, opened 2015-07-16
fixes #11139 - update puppet modules

stbenjam:
Trying to unpin theforeman-dns so we can update theforeman-puppet which contains the fix for this bug.


stbenjam:
[test]

ehelms:
ACK

PR #242, opened 2015-07-20
Fixes #11160 : Increase the number of times the scripts attempts to c…

bkearney:
…all cpinit before failing.

The prevailing thought is that this is due to lack of entropy on vms. If that is trye, all that can be done
is wait longer. This bumps up the script to attempt to connect 40 times, waiting 20 seconds between attempts.
This should give 800 seconds for the machine to get entropy

### DIFF
--- a/modules/candlepin/manifests/service.pp
+++ b/modules/candlepin/manifests/service.pp
@@ -12,7 +12,7 @@ class candlepin::service{
   if $candlepin::run_init == true {
     exec { 'cpinit':
       # tomcat startup is slow - try multiple times (the initialization service is idempotent)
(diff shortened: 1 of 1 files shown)


stbenjam:
APJ

ehelms:
NACK - you are making a change to a module directly instead of updating the module itself -- https://github.com/Katello/katello-installer#module-bugs

stbenjam:
Oh man, sorry, didn't notice that.  stbenjam--

bkearney:
Replaced by https://github.com/Katello/puppet-candlepin/pull/23

PR #245, opened 2015-07-22
fixes #11190 - load modules at build time, instead

stbenjam:
Jenkins needs changes of course to support this. I've taken parts of foreman-installers solution.  

`rake pkg:generate_source` generates the tarball for SRPM.


ehelms:
I had hoped we could avoid this change until installer scenario work was completed since that may shape how things look anyway. However, if you really want this change, consider it in light of the overall packaging change we are going to make @jlsherrill and myself are planning to work in earnest on that this week.

stbenjam:
I think this doesn't affect any packaging work. Or does it?

I do rely on katello-installer.spec, but I can use Foreman's solution of a VERSION file most likely, and I can even do that now if you want.  https://github.com/theforeman/foreman-installer/blob/develop/VERSION

The scenario work will certainly matter, but we're still going to have some structure where we have librarian and our modules, so this is still relevant.  Besides, I don't even know when the implementation phase is going to start

There's some advantages here that makes me really want this:

1) We can stop the double PR madness
2) I can stop ACK'ing direct PR's to this repo :grin: 
3) We can stop the double PR madness

The changes to jenkins *should* be relatively minor, too, and can be made backwards compatible (if Rakefile exists, do it the new way, otherwise do it the old way)

ehelms:
This will put more of a burden on our nightly builds in so much as that is where changes to the installer will get tested because we don't have extensive (or even basic in some cases) tests for all of our modules. I suppose we can monitor to see how much of an impact this has over time, we'll just need to ensure we are fixing nightly builds quickly since we will get a new build of the installer almost every day. It would be good for us to spend time to ensure some basic tests for all our modules as part of this.

ehelms:
One other question/thought, should we hook up all of our puppet modules to our prprocessor since we'd be fixing 'installer' errors on the individual modules and no longer tracking them against the installer itself?

ekohl:
I certainly like how this change has worked out on the foreman side. In general there are very few changes to the actual foreman-installer now. You really get continuous integration with all the benefits and downsides. I'm sure you asked @domcleal for his experience on this.

stbenjam:
Right, so we lose BATS against individual puppet module changes, because they're pulled from the repos every day, and tested on a nightly basis. 

But this means that Foreman's and Puppetforge modules get tested every night too.  We get insight into incompatibility much, much sooner for those.

It increases the chances of nightly build failures, and we have enough of those already.  Will it cause a lot more? I'm not sure. 

I don't think I want to run BATS against every Katello puppet change, it's really heavy and  I like Travis.

stbenjam:
You are right though, if you would prefer, we can hold off on this until we consider http://projects.theforeman.org/issues/11192 complete (better puppet test coverage).

ekohl:
Right now I'm quite content with how it works in foreman, but you maintain it so feel free to decide.

ehelms:
@stbenjam could you note which modules lack even a basic set of tests? We could split them up, add that at least prior to this.

stbenjam:
For manifests, just gutterball and elasticsearch. Elasticsearch lacks tests, but we're abandoning it eventually, right?

Maybe we can do something for common and service_wait too to test the custom types/providers/parser methods.

ekohl:
https://github.com/Katello/puppet-gutterball/pull/15 adds minimal tests to gutterball.

stbenjam:
You rock @ekohl 

puppet-common spec: https://github.com/Katello/puppet-common/pull/7

ehelms:
@stbenjam I don't think the Rakefile is needed. We figured out a way to use github tarball generation.

stbenjam:
@ehelms How? You still have to run librarian somehow.  And what if you want to do a scratch build from your own box?

ehelms:
Oh right, good point. That complicates things as our goal had been to avoid having to store tarballs somewhere and I think this setup will require that. @jlsherrill any thoughts since you came up with the github idea?

stbenjam:
Can it just be taken as an artifact from jenkins? How does foreman's job do it?

ehelms:
For jenkins yes - thats how it works even for rubygem-katello with the new structure. However, if you are building it for a given release, where you need to include or point at the source, I am less sure given that setup sources script doesn't know about grabbing a tarball, unwrapping it, running a rake task, tarballing it up again, etc.

jlsherrill:
One option would be that at release time, someone actually runs librarian puppet and checks that into some branch (possibly a release branch) and tags it.  It would be a manual process (possibly one we could automate) but would keep it consistent with our other builds.

stbenjam:
What's foreman-installer do? Seems we could do the same thing, no? 

Running librarian and committing it to git really stinks, I'd like to avoid that.

ekohl:
What @domcleal does is branch for every minor version (like https://github.com/theforeman/foreman-installer/commits/1.9-stable) and then pin the versions (like https://github.com/theforeman/foreman-installer/commit/ba4f3aa500a1f1c2f0e4ff77cf1068703d28fd4a). I believe the tarballs are stored in git annex, but I don't know the details there.

domcleal:
We generate tarballs in Jenkins as build artifacts for nightly releases (and use a tito builder to pull them in), and publish tarballs for releases (which git-annex fetches for RPMs).

stbenjam:
>  publish tarballs for releases

Publish where? On the web somewhere?

jlsherrill:
http://downloads.theforeman.org/foreman-installer/

is where they get published.

stbenjam:
So for all the other stuff Katello's building out of the github tarball?

We can commit it to the branch, as long as it stays out of master, which to me is the most important thing.

jlsherrill:
I was only suggesting pushing to git for releases (RC's and the main release).  Nightlies would still use jenkins artifacts.  

Personally It would be nice to have the releases in Git just to browse them (but I'd consider that a nice side effect).  I'm okay with storing them somewhere as well, I would assume we'd need a release jenkins job to do that (and somewhere to store them).

stbenjam:
> I was only suggesting pushing to git for releases (RC's and the main release). Nightlies would still use jenkins artifacts. 

That sounds OK to me :ok_hand:

ehelms:
I can't think of a better strategy so I'll work on crafting a jenkins job that takes advantage of this branch before we merge.

ehelms:
I built this job - http://ci.theforeman.org/job/release_build_katello_installer/configure which clones and generates the tarball using the rake task (its currently pointed here). I tested that by linking it up to the pipeline (currently only building a scratch build) which produced http://koji.katello.org/koji/taskinfo?taskID=316475 so from a nightly perspective I am good with this change.

ehelms:
I suppose as well, we'll either have to update the CI test job for the installer to account for this change or re-think our testing strategy for what is present here to more of a unit style test writing. Thoughts?

stbenjam:
> I suppose as well, we'll either have to update the CI test job for the installer to account for this change or re-think our testing strategy for what is present here to more of a unit style test writing. Thoughts?

My $0.02

The number of PR's is going to go down significantly, and we're going to be testing the installer as 
part of the pipeline.

I see the two most common PR's being changing Puppetfile and changing hooks.

For the former, it might be worthwhile to test that librarian can resolve all the dependencies to make sure we don't add something wrong to the Puppetfile.

For the latter, we don't test anyway, and I'm not sure we should invest a ton right now if we're going pluginfiy the installer at some point - but then we might want to look at testing upgrades in BATS or something?

stbenjam:
> For the former, it might be worthwhile to test that librarian can resolve all the dependencies to make sure we don't add something wrong to the Puppetfile.

And this one could probably be sent to Travis, too.  Take some load off poor jenkins

ehelms:
I think I would be fine with that -- want to include the travis config with this PR or separate?

stbenjam:
I can do it here

stbenjam:
Travis is up, it just builds the tarball, but it does indeed error out if there's something like dependency errors

ehelms:
Awesome. @jlsherrill any final thoughts or comments?

jlsherrill:
I'm good with all of this, APJ from me

ehelms:
ACK for me as well, I'll detach the PR job once this is merged and properly enable the pipeline job.

stbenjam:
Thanks :tada:

PR #246, opened 2015-07-22
refs #10716 - start httpd after migrating pulp

stbenjam:


hooks/pre/30-upgrade.rb
```diff
@@ -65,8 +65,8 @@ def fail_and_exit(message)
 
   if Kafo::Helpers.module_enabled?(@kafo, 'katello') || @kafo.param('capsule', 'pulp').value
     upgrade_step :start_mongo
-    upgrade_step :start_httpd
     upgrade_step :migrate_pulp
+    upgrade_step :start_httpd
```
ehelms: Do we restart other pulp services? Looking at the Pulp upgrade instructions we should be restarting workers/resource_manager/celerybeat after migrating as well.

stbenjam: They are stopped at this point.  See L64.  We only start httpd to be able to access the repos for a self-registered upgrade.  The other services will be started again during the upgrade.

The problem here was starting httpd for the repos before the migration, causes errors in the logs if you're missing things like docker_image in the database but have the pulp plugin installed.

stbenjam: Which made me realize another problem, we can't exclude httpd from the service stop, we have to have everything but mongo stopped for the migration.  So, I updated that.

ehelms:
APJ

stbenjam:
Thanks

PR #247, opened 2015-07-22
refs #10675 - templates off by default for upgrade, too

stbenjam:
I'm like 0 for 2 today.  https://github.com/Katello/katello-installer/pull/243 was not a complete fix. Templates would still get enabled on an upgrade, we need a migration to add it to existing answers.


ehelms:
ACK

PR #253, opened 2015-08-11
refs #11190 - add instructions for installer from git

stbenjam:


README.md
```diff
@@ -107,6 +107,17 @@ Install without RVM
 katello-devel-installer --katello-use-rvm false
 ```
 
+### From Git
+
+If you're working off the master branch, use Librarian to pull in all the modules to your local checkout:
```
ehelms: Could we include installation of librarian-puppet and puppet ruby > 2.0 since that isn't obvious?

stbenjam: There's a Gemfile now

stbenjam: Is that enough? Or should I specifically mention the requirements?

ehelms:
Oh, good call, LGTM! Could you reference the removal issue # so we aren't breaking protocol too much?

stbenjam:
Updated

ehelms:
ACK

ehelms:
ACK

PR #254, opened 2015-08-12
Fixes #11334: Change to gzip to mimic Github's compression

ehelms:
This also moves the compression format over to gzip to follow the
style that github uses since that is where we grab/store our archives.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -30,7 +30,7 @@ namespace :pkg do
 
     `git archive --prefix=katello-installer-#{version}/ HEAD > #{PKGDIR}/katello-installer-#{version}.tar`
     `tar --concatenate --file=#{PKGDIR}/katello-installer-#{version}.tar #{BUILDDIR}/modules.tar`
-    `bzip2 -9 #{PKGDIR}/katello-installer-#{version}.tar`
+    `gzip -9 #{PKGDIR}/katello-installer-#{version}.tar`
   end
 end
 


ehelms:
I could have the Jenkins job do this "conversion" for now, but I felt like it made more sense to capture the behavior here since we expect this to properly generate archives for use with packaging?

Rakefile
```diff
@@ -19,18 +19,21 @@ end
...
+      `tar -cf #{BUILDDIR}/modules.tar --transform=s,^,#{archive_name}/, modules/`
     end
 
-    `git archive --prefix=katello-installer-#{version}/ HEAD > #{PKGDIR}/katello-installer-#{version}.tar`
-    `tar --concatenate --file=#{PKGDIR}/katello-installer-#{version}.tar #{BUILDDIR}/modules.tar`
-    `bzip2 -9 #{PKGDIR}/katello-installer-#{version}.tar`
+    `git archive --prefix=#{archive_name}/ HEAD > #{PKGDIR}/#{archive_name}`
+    `tar --concatenate --file=#{PKGDIR}/#{archive_name} #{BUILDDIR}/modules.tar`
+    `gzip -9 #{PKGDIR}/#{archive_name}`
+    `mv #{PKGDIR}/#{archive_name}.gz #{PKGDIR}/#{archive_name}`
```
stbenjam: This ends up renaming the archive to just `.tar` instead of `.tar.gz`, was there a reason for that?

ehelms: I guess I could drop the tar altogether as well. The probably is mostly that github archives do not include the extensions '.tar.gz'

stbenjam: https://github.com/Katello/katello-installer/archive/master.tar.gz does

But even if it didn't, we should rename GitHub's not this one, a gzipped tar should end with `.tar.gz`.

ehelms:
Ok, I changed the intent of this after some more research into getting nightly builds with Jenkins to work -- see https://github.com/Katello/katello-packaging/pull/36

stbenjam:
ACK

PR #256, opened 2015-08-14
fixes #11356 - clean up puppetfile and add new reqs

stbenjam:
Also sorts ours alphabetically

### DIFF
--- a/Puppetfile
+++ b/Puppetfile
@@ -1,43 +1,37 @@
 forge 'https://forgeapi.puppetlabs.com'
 
-#################################################################
-# The Foreman part of Puppetfile (taken from foreman-installer #
-#################################################################
+##########################
+#        Foreman         #
+##########################
 
-# Dependencies
-mod 'puppetlabs-postgresql', :git => 'https://github.com/puppetlabs/puppetlabs-postgresql'
-mod 'puppetlabs-mysql', :git => 'https://github.com/puppetlabs/puppetlabs-mysql', :ref => '2.2.0'
-mod 'puppetlabs-firewall', :git => 'https://github.com/puppetlabs/puppetlabs-firewall', :ref => '1.0.0'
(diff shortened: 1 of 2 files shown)


ehelms:
If we are trying to publish our modules regularly on puppetforge, what is the reason we still want to point at git?

stbenjam:
The workflow I had in mind is the same as Foreman, where we go against git for nightlies, and at release time, we do the pinning to the minor released versions.

Smaller bug fixes its fine to go ahead and release to the forge, but if we're doing something substantial to the core modules I think we should feel the effect in the installer before releasing modules to the forge.

I did remove the `:git` for depedencies like mysql, mongodb, etc, to rely on released and stable versions from Puppetlabs.

stbenjam:
I mean, it doesn't have to be that way, I guess it depends how you see the modules.  We could release more often to the forge, but I figured nightly should be the picture of what's in git for everything, including the modules.

ehelms:
Fair enough - do we have a module release policy at the moment? I sorta like the idea of a release a commit given how in frequent they are, but if we are using our nightly installer to vet them I'd think we'd want some defined release cadence for them?

stbenjam:
Nope, so far I've released nearly every commit but I think its because we're in the RC cycle and there were a few little bugs.

IMHO, they should just be released as needed.

For example pulp 1.0 is on-going and we'll release that to the forge when we feel its ready (along with capsule + others using it).  We can't release every commit in that case.

It's also worthwhile for the releaser to spend some time thinking about what the changes were since the last release.  Because the process of releasing a module being something done "as needed" and not something we do everyday or twice a day, sends a signal to us to consider installer migrations, something we didn't have before.  More than once we had to deal with those problems much later when testing upgrades, perhaps even after RC's are in users hands.

ehelms:
Ok... I think I can live with that for now. ACK

PR #257, opened 2015-09-01
refs #11648 - ensure postgresql is started

stbenjam:


hooks/pre/30-upgrade.rb
```diff
@@ -1,17 +1,23 @@
+def start_service(service)
+  status = Kafo::Helpers.execute("service-wait #{service} status")
```
ehelms: Guessing katello-service doesn't have a start --only type of option such that we aren't relying on two service interfaces?

stbenjam: We could, I'll have to check. Mostly worried about what upstart does when you try to start an already started service, I think that's why I check the status first here.  systemd will return 0, but all of the old init scripts might not play so nicely.

stbenjam:
Updated to use @bbuckingham's `--only`

ehelms:
ACK

PR #258, opened 2015-09-02
Fixes #11666: Cleanup _build and pkg directories between runs

ehelms:
Successive runs of the rake task that generates package tarballs
did not remove or overwrite the previous builds. In a local or CI
environment, the tarball built could unknowingly result in no
updates and a stale package.


ehelms:
I noticed this issue while trying to update our nightly pipeline for the TFM SCL. I kept seeing the same systest error because it was not re-generating a new tarball with the latest and greatest.

stbenjam:
ACK

PR #260, opened 2015-09-08
fixes #11737 - dispatch router connect to qpid on localhost

stbenjam:
Removes the FQDN's for the AMQP hosts so that the new defaults are used (localhost).

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -20,3 +20,14 @@ Style/GuardClause:
 
 Style/SpaceInsideHashLiteralBraces:
   Enabled: false
+
+Metrics/AbcSize:
+  Enabled: false
+
+Metrics/CyclomaticComplexity:
+  Enabled : false
+
+Metrics/PerceivedComplexity:
+  Enabled: false
+
+
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -34,10 +34,6 @@ Lint/UselessAssignment:
 Metrics/BlockNesting:
   Max: 4
 
-# Offense count: 1
-Metrics/CyclomaticComplexity:
-  Max: 8
-
 # Offense count: 170
 # Configuration parameters: AllowHeredoc, AllowURI, URISchemes.
 # URISchemes: http, https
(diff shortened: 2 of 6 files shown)


stbenjam:
WIP, because capsule-installer will need it too

stbenjam:
@beav @ehelms This should address the update scenario.  To test:

- Build a normal Katello box
- Create vars.yaml (notice installer being skipped)

```yaml
---
foreman_repositories_use_koji: True
katello_repositories_use_koji: True
foreman_installer_module_prs: "katello/candlepin/44,katello/katello/79,katello/capsule/96,katello/certs/65"
foreman_installer_skip_installer: True
puppet_repositories_version: 4
```

- Run the playbook w/ the modules-pr branch checked out

```
ansible-playbook -v -l qpid-localhost playbooks/katello.yml --extra-vars "@vars.yaml"
```
- Download the files (sorry, TODO item for forklift) from the katello-installer pr and drop them into the right places

- Run foreman-installer

Result:
 qpid cert gets updated, config gets updated, and everything ends up happy

stbenjam:
I tried this today and it does work.

ehelms:
Working on finishing upgrade testing of this, while I do so, @stbenjam should we get some documentation started around upgrades while this is fresh?

stbenjam:
@ehelms I think there's nothing to do - Katello will Just Work :tm:, and for Capsule's we already ask users to regenerate certs:
  http://www.katello.org/docs/3.2/upgrade/capsule.html

Although we probably should only do that when it's needed (but will be for 3.3 since I expect that's when this will make it in right?)

ehelms:
@stbenjam hah, good find, who knew. So the server doesn't need a --certs-update-all?

stbenjam:
>  So the server doesn't need a --certs-update-all?

Nope, I added an upgrade hook this morning to do it automagically

ehelms:
@stbenjam will this hook run for every upgrade?

stbenjam:
Not anymore https://github.com/Katello/katello-installer/pull/408

ehelms:
I believe #408 only applies to post-upgrade steps and this is a pre-upgrade step.

stbenjam:
Right, will add that here

ehelms:
I think it is worth while for this to be separated. The pre-hook upgrade work could be useful stand-alone or as a backported commit.

stbenjam:
@ehelms https://github.com/Katello/katello-installer/pull/416

stbenjam:
I'll rebase this one once that's merged

stbenjam:
Rebased

ehelms:
ACK

PR #264, opened 2015-09-17
Updated cert check script for new installs since --update-certs is no…

chris1984:
…t needed for a first time run


ehelms:
Redmine issue please, and if you wouldn't mind updating your commit message to a shorter (~50 characters) and then a longer message following, see http://www.katello.org/developers/style/git_commit_messages.html

PR #265, opened 2015-09-17
Fixes #11856 - Updating katello-certs-check

chris1984:
updated script to take into account new installs, as the --update-certs flag is not needed for a 1st time run


chris1984:
Closed last PR because I am not sure what I did to get it into that state with 7 files and I was having issues resetting everything back to a fresh start. This one is correct that refs the upstream Redmine Issue.

PR #267, opened 2015-09-24
Fixes #10291 - removes elasticsearch

cfouant:
Goes with https://github.com/Katello/katello-packaging/pull/82 and https://github.com/Katello/puppet-katello/pull/85

### DIFF
--- a/hooks/post/30-upgrade.rb
+++ b/hooks/post/30-upgrade.rb
@@ -38,6 +38,18 @@ def import_subscriptions
   Kafo::Helpers.execute('foreman-rake katello:upgrades:2.4:import_subscriptions')
 end
 
+def remove_elasticsearch
+  if Kafo::Helpers.execute('rpm -q elasticsearch')
+    gems = ['ruby193-rubygem-tire', 'tfm-rubygem-tire', 'elasticsearch', 'sigar', 'snappy-java', 'lucene4-contrib', 'lucene4']
+    gems.each do |gem|
+      Kafo::Helpers.execute("rpm -e #{gem}")
+    end
+    message = "Elasticsearch has been removed as a dependency, the database files can be "\
+              "removed manually with #rm -rf /var/lib/elasticsearch"
+    Kafo::Helpers.log_and_say :info, message
+  end
+end
+
 def upgrade_step(step, options = {})
   noop = app_value(:noop) ? ' (noop)' : ''
   long_running = options[:long_running] ? ' (this may take a while) ' : ''
@@ -67,6 +79,7 @@ if app_value(:upgrade)
     upgrade_step :import_distributions, :long_running => true
     upgrade_step :import_puppet_modules, :long_running => true
     upgrade_step :import_subscriptions, :long_running => true
+    upgrade_step :remove_elasticsearch
   end
 
   if [0,2].include? @kafo.exit_code
--- a/Puppetfile
+++ b/Puppetfile
@@ -26,7 +26,6 @@ mod 'katello-capsule',          :git => 'https://github.com/Katello/puppet-capsu
 mod 'katello-certs',            :git => 'https://github.com/Katello/puppet-certs'
 mod 'katello-common',           :git => 'https://github.com/Katello/puppet-common'
 mod 'katello-crane',            :git => 'https://github.com/Katello/puppet-crane'
-mod 'katello-elasticsearch',    :git => 'https://github.com/Katello/puppet-elasticsearch'
 mod 'katello-gutterball',       :git => 'https://github.com/Katello/puppet-gutterball'
 mod 'katello-katello',          :git => 'https://github.com/Katello/puppet-katello'
 mod 'katello-pulp',             :git => 'https://github.com/Katello/puppet-pulp'
--- a/hooks/pre/10-reset_feature.rb
+++ b/hooks/pre/10-reset_feature.rb
@@ -7,7 +7,6 @@ def reset
     reset_database
     reset_candlepin
     reset_pulp
-    reset_elasticsearch
 
   else
     Kafo::KafoConfigure.logger.warn 'Katello not installed yet, can not drop database!'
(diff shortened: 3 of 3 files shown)


hooks/post/30-upgrade.rb
```diff
@@ -38,6 +38,11 @@ def import_subscriptions
   Kafo::Helpers.execute('foreman-rake katello:upgrades:2.4:import_subscriptions')
 end
 
+def remove_elasticsearch
+  Kafo::Helpers.execute('rpm -e elasticsearch lucene3 ruby193-rubygem-tire cigar') if system('rpm -q elasticsearch')
```
ehelms: This will need to remove tfm-rubygem-tire.

ehelms: Testing this locally, this didn't get executed probably due to what the `system()` call is returning. You may want to just use `Kafo::Helpers.execute` for this.

ehelms: I think the package is `sigar` and not `cigar`.

hooks/post/30-upgrade.rb
```diff
@@ -38,6 +38,11 @@ def import_subscriptions
...
+  Kafo::Helpers.execute('rm -rf /var/lib/elasticsearch')
```
jlsherrill: I'm kinda hesitant to blindly rm -rf this directory.  I'd rather we document that the user can do this.

hooks/post/30-upgrade.rb
jlsherrill: NACK, prefer to avoid --nodeps.  Should be able to pull it off removing them one by one in the right order

ehelms: Perhaps for some, but you can't remove elasticsearch without --nodeps because rubygem-katello requires it? I suppose if the updated RPM is installed first then it might work without it. I was not testing that scenario.

jlsherrill: right, it should work in that scenario which is the right scenario to test in :)

cfouant:
Testing the order of package removal to avoid needing the --nodeps option.

cfouant:
@ehelms @jlsherrill I don't have capabilities to change the labels on this PR, but I believe it's ready for retesting. This *should* be the correct order.

hooks/post/30-upgrade.rb
```diff
@@ -38,6 +38,14 @@ def import_subscriptions
   Kafo::Helpers.execute('foreman-rake katello:upgrades:2.4:import_subscriptions')
 end
 
+def remove_elasticsearch
+  gems = ['ruby193-rubygem-tire', 'tfm-rubygem-tire', 'elasticsearch', 'sigar', 'lucene3']
+  gems.each do |gem|
+    Kafo::Helpers.execute('rpm -e #{gem}')
+  end
+  echo "Remove elastic search files with #rm -rf /var/lib/elasticsearch"
```
cfouant: @jlsherrill - I wasn't sure if your previous comment meant to document during the upgrade, or in the docs themselves. Let me know what you think, thanks!

ehelms: I think this is a good spot for it, however, I'd give them more information here. Something explaining that Elasticsearch has been removed as a dependency but the user needs to manually remove the database files and that they can do so via your command listed.

jlsherrill: Our current upgrade scripts run every time --upgrade is used, so I'm not sure if that is the best spot.  I was more thinking on katello.org.

hooks/post/30-upgrade.rb
```diff
@@ -38,6 +38,15 @@ def import_subscriptions
   Kafo::Helpers.execute('foreman-rake katello:upgrades:2.4:import_subscriptions')
 end
 
+def remove_elasticsearch
+  gems = ['ruby193-rubygem-tire', 'tfm-rubygem-tire', 'elasticsearch', 'sigar', 'lucene3']
+  gems.each do |gem|
+    Kafo::Helpers.execute('rpm -e #{gem}')
+  end
+  echo "Elasticsearch has been removed as a dependency, the database files can be removed manually."
+  echo "Remove elastic search files with #rm -rf /var/lib/elasticsearch"
```
ehelms: We could check for it's existence and only put out the message if found? and/or if RPMs were removed?

Also, can't use echo here -- that's not a function in Ruby, better to use:

```
   Kafo::Helpers.log_and_say :info, message
```

jlsherrill: Works for me!

hooks/post/30-upgrade.rb
```diff
@@ -38,6 +38,15 @@ def import_subscriptions
   Kafo::Helpers.execute('foreman-rake katello:upgrades:2.4:import_subscriptions')
 end
 
+def remove_elasticsearch
+  gems = ['ruby193-rubygem-tire', 'tfm-rubygem-tire', 'elasticsearch', 'sigar', 'lucene3']
+  gems.each do |gem|
+    Kafo::Helpers.execute('rpm -e #{gem}')
```
ehelms: This needs to be double quotes to interpolate:

```
  Kafo::Helpers.execute("rpm -e #{gem}")
```

ehelms:
I still had snappy-java and lucene4 installed on my EL7 box.

hooks/post/30-upgrade.rb
ehelms: @jlsherrill should this use `yum remove` instead to ensure that dependencies get removed?

jlsherrill: yum remove will remove the package you request and everything that depends on it, i would be hesitant to do that personally as it could remove something unintentional.

cfouant:
@ehelms @jlsherrill can you look at the new list of gems and tell me if there are any others I should be including? Thanks!!!

jlsherrill:
I don't think lucene3 is needed?  Doesn't look like we've shipped it since before 2.0?

cfouant:
@ehelms @jlsherrill also, I'm pretty sure my method for checking that elasticsearch exists is incorrect, but not sure what the correct conditional would be.

jlsherrill:
@cfouant i'm okay with that check, looks good to me.  Will let ehelms do final review

ehelms:
ACK

cfouant:
Thanks @ehelms @jlsherrill !!! I will wait on one of you for the merge.

ehelms:
Thanks @cfouant !

PR #270, opened 2015-10-15
Pin modules for Katello 2.4

stbenjam:


stbenjam:
Removed the modules

jlsherrill:
ACK, thanks @stbenjam !

PR #271, opened 2015-10-26
adding modules for 2.4 release

jlsherrill:
these actually are needed so that we have somewhere to point to for the release process.


stbenjam:
ACK

PR #283, opened 2015-12-16
fixes #10533 - update answers and config to decouple foreman_proxy from capsule

bbuckingham:
This commit contains some initial changes to support the puppet-capsule
changes that decouple puppet-foreman_proxy from the puppet-capsule.
(Ref: https://github.com/Katello/puppet-capsule/pull/64)

The following is an example to install katello using this configuration:
```
katello-installer
```

The following is an example to install capsule using this configuration:
```
capsule-installer --register-in-foreman "true"\
                  --registered-proxy-url "https://capsule.example.com:9090"\
                  --oauth-consumer-key "pvUutwvPU3Rni5RWqisL6Tx4fAF5yDSa"\
                  --oauth-consumer-secret "JTGnd3fx6GB4uki5LtNPFU6cCSq9NrFk"\
                  --foreman-base-url "https://katello.example.com"\
                  --trusted-hosts "katello.example.com"\
                  --trusted-hosts "capsule.example.com"\
                  --parent-fqdn "katello.example.com"\
                  --pulp-oauth-secret "3HNmounKP6mbQYthkvkw2GXV5toxeFsu"\
                  --certs-tar "~/capsule.example.com-certs.tar"\
                  --puppet "true"\
                  --puppetca "true"\
                  --enable-foreman-proxy-plugin-pulp\
                  --enabled "false"\
                  --pulpnode-enabled "true"
```

### DIFF
--- a/config/answers.capsule-installer.yaml
+++ b/config/answers.capsule-installer.yaml
@@ -3,3 +3,18 @@ certs:
   generate: false
   deploy: true
 capsule: true
+foreman_proxy:
+  custom_repo: true
+  http: true
+  ssl_port: '9090'
+  templates: true
+  tftp: false
+  ssl_ca: /etc/foreman-proxy/ssl_ca.pem
+  ssl_cert: /etc/foreman-proxy/ssl_cert.pem
+  ssl_key: /etc/foreman-proxy/ssl_key.pem
+  foreman_ssl_ca: /etc/foreman-proxy/foreman_ssl_ca.pem
+  foreman_ssl_cert: /etc/foreman-proxy/foreman_ssl_cert.pem
+  foreman_ssl_key: /etc/foreman-proxy/foreman_ssl_key.pem
+foreman_proxy::plugin::pulp:
+  enabled: false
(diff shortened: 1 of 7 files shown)


config/answers.capsule-installer.yaml
```diff
@@ -3,3 +3,21 @@ certs:
   generate: false
   deploy: true
 capsule: true
+foreman_proxy:
+  custom_repo: true
+  puppetca: true
+  http: true
+  http_port: '8000'
+  ssl_port: '9090'
+  register_in_foreman: true
```
stbenjam: I think we can remove anything that's already default by Foreman: https://github.com/theforeman/puppet-foreman_proxy/blob/master/manifests/params.pp

stbenjam:
We're going to need some migration to convert old capsule answers to the equivalent foreman proxy settings

ehelms:
ACK pending rebase -- I'm good to merge the puppet-capsule change and this. I ran a nightly install test against it and content bats tests which all passed.

PR #284, opened 2015-12-18
Fixes #12881 - Add an installer check about the LANG environment vari…

bkearney:
…able

### DIFF
--- a/checks/lang.rb
+++ b/checks/lang.rb
@@ -1,0 +1,10 @@
+#!/usr/bin/env ruby
+
+INVALID_LANG = %q(The LANG envrionment variable should not be set to C)
+
+def error_exit(message, code)
+  $stderr.puts message
+  exit code
+end
+
+error_exit(INVALID_LANG, 1)  if ENV["LANG"] == "C"


checks/lang.rb
```diff
@@ -0,0 +1,14 @@
...
+
+INVALID_LANG = %q(The LANG envrionment variable should not be set to C)
+
+def error_exit(message, code)
+  $stderr.puts message
+  exit code
+end
+
+begin
+    error_exit(INVALID_LANG, 1)  if ENV["LANG"] == "C"
```
ehelms: 2 space indent

checks/lang.rb
```diff
@@ -0,0 +1,14 @@
...
+rescue
+    # Eat the exception and continue
```
ehelms: What exception might be thrown here? Looks like the code in the begin statement is a fairly simple check and function call

bkearney: Likely none, but this seemed to be the pattern for the other checks. I assumed the logic was these should never fail

ehelms:
I'd say that begin/end is only the format for one check, e.g. https://github.com/Katello/katello-installer/blob/master/checks/hostname.rb#L41-L49

bkearney:
@ehelms removed the extraneous begin/end

ehelms:
ACK

PR #287, opened 2016-01-18
Fixes #13250 - Remove capsule 9090 url from install

chris1984:
removed url that gives fqdn:9090 where customers try to visit and get an error.


beav:
needs rebase but the change itself looks fine

PR #290, opened 2016-01-26
Fixes #12550, #10533 - Convert capsule installer to scenarios

mbacovsky:
This PR should replace https://github.com/Katello/katello-installer/pull/283.
Included are migrations from old configuration which are prevented with the PR above.

This PR also removes the foreman related modules as they are loaded from foreman-installer.


ehelms:
Looks like this will need a rebase. 

This section of user output https://github.com/Katello/katello-installer/pull/290/files#diff-d9b1058a236a3d65bd0c4fbc0b60f640R28 will need updating. You can do it here or in a subsequent PR once we have things ironed out more.

ehelms:
Now green after my updates to puppet-candlepin (https://github.com/Katello/puppet-candlepin/pull/37)

mbacovsky:
@ehelms, thanks a lot!

config/capsule.migrations/160120212700-foreman-proxy.rb
```diff
@@ -0,0 +1,64 @@
...
+    move('tftp', false)
+    move('tftp_syslinux_root')
+    move('tftp_syslinux_files')
+    move('tftp_root', '/var/lib/tftpboot')
+    move('tftp_dirs', ['/var/lib/tftpboot/pxelinux.cfg', '/var/lib/tftpboot/boot'])
+    move('tftp_servername')
+
+    move('foreman_proxy_port', '9090', 'http_port')
+    move('foreman_proxy_http', true, 'http')
+    move('foreman_proxy_http_port', '8000', 'ssl_port')
```
ehelms: This doesn't appear to work on non-expanded answers files. For example, if using this as is, `ssl_port` is not defined in the answers file but is defined within the old puppet-capsule. So when this generates an answers file on greenfield installations, the ssl_port gets set to 8000 instead of 9090.

mbacovsky: Could it be that there is just a bug in the migration? Shouldn't the ports be:
```ruby
    move('foreman_proxy_port', '9090', 'ssl_port')
    move('foreman_proxy_http', true, 'http')
    move('foreman_proxy_http_port', '8000', 'http_port')
```
?, note the `ssl_port` and `http_port` exchange.

ehelms: Hrmm, yes -- I'll test that

config/capsule.migrations/160120212700-foreman-proxy.rb
```diff
@@ -0,0 +1,64 @@
...
+    move('foreman_oauth_effective_user', 'admin', 'oauth_effective_user')
+    move('foreman_oauth_key', 'oauth', 'oauth_consumer_key')
+    move('foreman_oauth_secret', 'secret', 'oauth_consumer_secret')
```
ehelms: I think these two oauth settings should be omitted here.

mbacovsky:
- fixed the capsule concerns (updates are in separate commit, needs squash before it is merged)
- cherry-picked content of #291 to have evrything in one PR for better testing

mbacovsky:
@ehelms it seems these migrations were not released yet: https://github.com/Katello/katello-installer/commit/f5c81ee1b16c3e1a6443b138e54bc4ff2c418439. Shouldn't we add this migration in `installer-scenarios.d/katello.migrations/`?

ehelms:
1.10/2.4 were released with tfm- pacakges (just not Rails 4 and Ruby 2.2) so those migrations should have been applied.

mbacovsky:
I hit an issue while switching scenario form Foreman to Katello. 

During scenario switch installer needs to parse modules for both scenarios and shares the parsed classes on one place. Katello scenario ships its moduels with deps so we end up with e.g. foreman module in both the foreman and the katello installers. Puppet is then confused as it finds both. 

Workaround would be to change ordering of module dirs in katello scenario (not tested) but I'd prefer not to ship duplicate modules. I patched the Rakefile to remove the duplicities. Con is that to run pkg:generate_source you have to have foreman-installer modules somewhere. By default it expects foreman-installer is installed and you can run just `rake pkg:generate_source` as usual or you have local git checkout with modules built and then you can run rake with `rake pkg:generate_source FOREMAN_MODULES_DIR=../foreman-installer/_build/modules`.

I'm thinking about implementing third option which is that rake clones and build foreman modules somewhere in temp. 

New katello-installer built without dupe modules is: 
http://koji.katello.org/koji/taskinfo?taskID=417040

bbuckingham:
I am not sure if this is the proper PR to comment on; however, I am testing with the PR when observing the warnings below.  Note: the config is based on the email thread on theforeman-dev which involves scratch RPMs and katello nightly.

```
# foreman-installer --list-scenarios
WARN: Unresolved specs during Gem::Specification.reset:
      ansi (>= 0)
      logging (< 3.0.0)
      powerbar (>= 0)
WARN: Clearing out unresolved specs.
Please report a bug if this causes problems.
Available scenarios
  Foreman (use: --scenario foreman)
        Default installation of Foreman
  Capsule (use: --scenario capsule)
        Install a stand-alone Capsule.
  Katello (use: --scenario katello)
        Install Foreman with Katello
```

bin/capsule-certs-generate
```diff
@@ -1,10 +1,54 @@
...
+answers_file = temp_file('capsule-certs-answers', ANSWERS_CONTENT)
+
+CONFIG_CONTENT = <<EOF
+:log_level: :debug
+:log_dir: /var/log
+:log_name: capsule-certs-generate.log
+:no_prefix: true
+:answer_file: #{answers_file.path}
+:installer_dir: #{INSTALLER_DIR}
+:module_dirs: ["#{INSTALLER_DIR}/modules"]
```
bbuckingham: We need to also include the modules from foreman (e.g. /usr/share/foreman-installer/modules); otherwise, running capsules-certs-generate will generate an error similar to:

```
[ERROR 2016-02-09T12:32:21 main]  Unknown function cache_data at /usr/share/foreman-installer-katello/modules/certs/manifests/init.pp:127 on node katello-scenario-katello.example.com
```

config/capsule.yaml
```diff
@@ -0,0 +1,16 @@
...
+  :log_name: capsule.log
+  :log_level: DEBUG
+  :no_prefix: true
+  :answer_file: "./config/capsule-answers.yaml"
+  :installer_dir: "."
+  :default_values_dir: "/tmp"
+  :module_dirs: ["/usr/share/foreman/modules", "./modules"]
+  :hook_dirs: ["./hooks"]
+  :colors: true
+  :order:
```
bbuckingham: It looks like we are going to need to specify the order as the following:

```
    - certs
    - foreman_proxy
    - "foreman_proxy::plugin::pulp"
    - capsule
```

config/capsule-answers.yaml
```diff
@@ -0,0 +1,5 @@
+---
+certs:
+  generate: false
+  deploy: true
+capsule: true
```
bbuckingham: Today, capsule-certs-generate indicates to include **--puppet "true"**, should we continue to include that or simply update the answers (or migration) to ensure the following, so that the user doesn't need to specify it:

```
capsule:
  puppet: true
```

bbuckingham:
Testing with the changes in this PR, capsule-certs-generate will create output similar to the following (as defined in 10-post_install.rb):

```
  capsule-installer --parent-fqdn          "katello-scenario-katello.example.com"\
                    --register-in-foreman  "true"\
                    --foreman-oauth-key    "AdfxMBXAnaVXhrLAY5wCfb4Tt72WYchu"\
                    --foreman-oauth-secret "yUQsV3ESjBDuyQDxc2sB7asAsjeHWE49"\
                    --pulp-oauth-secret    "FmV7ExFLmgMtr28W7Rovs5KY7fYtU5fs"\
                    --certs-tar            "~/katello-scenario-capsule.example.com-certs.tar"\
                    --puppet               "true"\
                    --puppetca             "true"\
                    --pulp                 "true"
```

With the scenario changes introduced, we will need to update this output.  In my test, I ran with the following:
```
foreman-installer --scenario capsule \
                  --parent-fqdn           "katello-scenario-katello.example.com"\
                  --registered-proxy-url  "https://katello-scenario-capsule.example.com:9090"\
                  --oauth-consumer-key    "AdfxMBXAnaVXhrLAY5wCfb4Tt72WYchu"\
                  --oauth-consumer-secret "yUQsV3ESjBDuyQDxc2sB7asAsjeHWE49"\
                  --foreman-base-url      "https://katello-scenario-katello.example.com"\
                  --trusted-hosts         "katello-scenario-katello.example.com"\
                  --trusted-hosts         "katello-scenario-capsule.example.com"\
                  --pulp-oauth-secret     "FmV7ExFLmgMtr28W7Rovs5KY7fYtU5fs"\
                  --certs-tar             "~/katello-scenario-capsule.example.com-certs.tar"\
                  --puppet                "true"
```

Note: per an earlier comment, we could omit **--puppet "true"**, if we add it to the answers/migration.

config/capsule-answers.yaml
mbacovsky: another option is to set new default in `puppet-capsule`. If there are reasons against i'd go with the 
migration.

mbacovsky:
Regarding the gem warnings I'm suspecting `katello-deploy` installing kafo and deps from rubygems. I'll have to test to confirm.

config/capsule-answers.yaml
ehelms: I personally don't think the 'capsule' should enforce puppet to be setup by default - if the smart proxy doesn't do this by default neither should we.

bbuckingham: Why do we have capsule-certs-generate include puppet=true, if we don't feel that the majority of installations would be including it?  Would not including puppet affect pulp's puppet support?  (I assume not, but curious.)

bbuckingham: Note: there is an existing issue related to having puppet disabled : http://projects.theforeman.org/issues/11197 . I am surprised that more folks have not hit it.

mbacovsky:
I updated the PR, the changes are in separate commits.
- katello-devel converted to scenarios
- fixed module path in capsule-certs-generate
- added migration with puppet = true by default (can be removed if we decide to)
- updated modules ordering by overriding it in migration. It is safer for more cases. It can break existing config if users have own ordering/own modules in though. If that is an issue I'll try to find out something better.

mbacovsky:
I've added new commit containing changes necessary for katello-installer being standalone.
- new binary `katello-installer`
- log dir is `/var/log/katello-installer`
- modules are expected in `katello-installer/modules` and include also foreman modules

I've added it as a new commit and not new PR to preserve the history and cherry pick for now.

config/capsule-answers.yaml
bbuckingham: @mbacovsky, @ehelms, This is to follow-up to our discussion the other day.  The question was whether or not it is possible to sync a puppet repo (which is using the puppet install distributor) to a capsule, if puppet was not configured on the capsule.  I was finally able to run a test with upstream katello+capsule and the following scenario does generate an error to the user:
```
1. configure the capsule to sync the 'dev' environment
2. create a puppet repo & upload a module to it
3. create a content view, add the puppet module to it & publish to library
4. promote that content view to 'dev', which will trigger the content to be synced to the capsule
```

In the above scenario, the sync to the capsule will fail, since /etc/puppet/environments is non-existent.  The error indicates the following:
```
PLP0034: The distributor Default_Organization-dev-puppet_view indicated a failed response when publishing repository Default_Organization-dev-puppet_view.
```

Puppetfile
```diff
@@ -1,26 +1,23 @@
...
+
+# Top-level modules
+mod 'theforeman/foreman',       :git => 'https://github.com/theforeman/puppet-foreman'
+mod 'theforeman/foreman_proxy', :git => 'https://github.com/theforeman/puppet-foreman_proxy'
+mod 'theforeman/puppet',        :git => 'https://github.com/theforeman/puppet-puppet'
+
 
-# Katello specific modules
+# Katello modules
 mod 'katello-candlepin',        :git => 'https://github.com/Katello/puppet-candlepin'
```
ekohl: Make these consistent and also write them with a slash instead of a dash?

Rakefile
```diff
@@ -11,11 +13,35 @@ file PKGDIR do
...
+
+desc 'Remove foreman modules from $BUILDDIR/modules'
+task :prune_foreman_modules => [] do
+  begin
+    # raise "No foreman modules found in #{FOREMAN_MODULES_DIR}"
+    if Dir["#{FOREMAN_MODULES_DIR}/*"].empty?
+      temp_dir = Dir.mktmpdir
+      Dir.chdir(temp_dir) do
+        puts 'Downloading foreman modules...'
+        `git clone https://github.com/theforeman/foreman-installer.git`
```
ekohl: Do you want a branch here as well for releases?

ehelms:
One interesting difference we will run into with this is on a server install there will be two scenarios presented to the user: foreman and katello. On what should be a capsule install, users will be presented with: foreman and capsule.

bbuckingham:
@ehelms, perhaps only 'capsule' should be presented there for a capsule install.  On the test that I did yesterday with the plan B katello packages, I only installed katello-installer-capsule, so it was the only scenario presented and the install was clean.

config/capsule.migrations/160120212700-foreman-proxy.rb
```diff
@@ -0,0 +1,66 @@
+# Decouple puppet-foreman_proxy form puppet-capsule
```
ekohl: Nitpick: s/form/from/

Rakefile
mbacovsky: Good point, it didn't came to my mind we would want to build against older foreman. I'll add variable to handle this to the rake file

mbacovsky:
@ehelms, @bbuckingham re multiple scenarios. With plan A this is true. This is just potential UX issue though. From this perspective it may make better sense to have all katello scenarios (katello, capsule, katello-devel) in one rpm and display all the scenarios to the user everytime.

mbacovsky:
Updated, squashed, all comments fixed.

bbuckingham:
We do need a minor update the capsule-certs-generate output.

On my install, I saw the following:
```
  foreman-installer --scenario capsule\
                    --parent-fqdn                         "katello-scenario-katello.example.com"\
                    --register-in-foreman                 "true"\
                    --foreman-proxy-oauth-consumer-key    "JWWbB4CDVHVYScAgTopBi9PxmiPhkJe6"\
                    --foreman-proxy-oauth-consumer-secret "MCFBM49sDbWkcSvRJtv5o9bJUvTJAvmJ"\
                    --capsule-pulp-oauth-secret           "kW3N7FhmSLSJU35wDT2PJTMR7rgtNXh6"\
                    --certs-tar                           "~/katello-scenario-capsule.example.com-certs.tar"
```

But, I needed the following:
(Note: removal of 'foreman-proxy-' and 'capsule-' from a few params.)
```
  foreman-installer --scenario capsule\
                    --parent-fqdn           "katello-scenario-katello.example.com"\
                    --register-in-foreman   "true"\
                    --oauth-consumer-key    "JWWbB4CDVHVYScAgTopBi9PxmiPhkJe6"\
                    --oauth-consumer-secret "MCFBM49sDbWkcSvRJtv5o9bJUvTJAvmJ"\
                    --pulp-oauth-secret     "kW3N7FhmSLSJU35wDT2PJTMR7rgtNXh6"\
                    --certs-tar             "~/katello-scenario-capsule.example.com-certs.tar"
```

mbacovsky:
That is interesting as foreman-installer --scenario capsule --help list them with the prefix...

config/capsule-answers.yaml
mbacovsky: @bbuckingham Thanks for testing it. Does it mean we want to keep puppet: true in by default?

bbuckingham: @mbacovsky, I lean towards having it default to 'true' since we always included it in the 'capsule-certs-generate' and there are bugs that exist when it is not enabled; however, I am open to other opinions.  @ehelms, thoughts?

ehelms: We can err to true for now and relax it later.

ehelms:
I'm satisfied with my testing of this, it would be nice to get the documentation for the capsule output as @bbuckingham indicates but its not a blocker for merging this.

mbacovsky:
On my fresh install I can now see the oauth params NOT prefixed as well. I'll update the certs generate doc doc to reflect that.

mbacovsky:
Now the updated version looks like:
```

  1. Ensure that the katello-installer-capsule package is installed on the system.
  2. Copy ~/katello-scenario-capsule.example.com-certs.tar to the system katello-scenario-capsule.example.com
  3. Run the following commands on the capsule (possibly with the customized
     parameters, see foreman-installer --scenario capsule --help and

  yum -y localinstall http://katello-scenario-katello.example.com/pub/katello-ca-consumer-latest.noarch.rpm
  subscription-manager register --org "Default_Organization"
  foreman-installer --scenario capsule\
                    --parent-fqdn                         "katello-scenario-katello.example.com"\
                    --register-in-foreman                 "true"\
                    --foreman-base-url                    "https://katello-scenario-katello.example.com"\
                    --trusted-hosts                       "katello-scenario-katello.example.com"\
                    --trusted-hosts                       "katello-scenario-capsule.example.com"\
                    --oauth-consumer-key                  "UVrAZfMaCfBiiWejoUVLYCZHT2xhzuFV"\
                    --oauth-consumer-secret               "ZhH8p7M577ttNU3WmUGWASag3JeXKgUX"\
                    --pulp-oauth-secret                   "TPk42MYZ42nAE3rZvyLBh7Lxob3nEUi8"\
                    --certs-tar                           "~/katello-scenario-capsule.example.com-certs.tar"
  The full log is at /var/log/capsule-certs-generate.log
```

bbuckingham:
@mbacovsky, excellent!  Thanks for the updates!

bbuckingham:
@mbacovsky, @ehelms, tested with the latest builds and https://github.com/Katello/katello-packaging/pull/151.

All is working well; therefore, ACK!!

PR #294, opened 2016-02-02
Fixes #13520 - Use correct mongod.conf path in 30-upgrade hook for EL7

Thubo:
See http://projects.theforeman.org/issues/13520


ehelms:
ACK - thanks @Thubo !

PR #295, opened 2016-02-10
refs #13454 - add docker v1 repository removal into upgrade actions

komidore64:
this goes along with Katello/katello#5744

### DIFF
--- a/hooks/post/30-upgrade.rb
+++ b/hooks/post/30-upgrade.rb
@@ -50,6 +50,10 @@ def remove_elasticsearch
   end
 end
 
+def remove_docker_v1_content
+  Kafo::Helpers.execute('foreman-rake katello:upgrades:3.0:delete_docker_v1_content')
+end
+
 def upgrade_step(step, options = {})
   noop = app_value(:noop) ? ' (noop)' : ''
   long_running = options[:long_running] ? ' (this may take a while) ' : ''
@@ -80,6 +84,7 @@ if app_value(:upgrade)
     upgrade_step :import_puppet_modules, :long_running => true
     upgrade_step :import_subscriptions, :long_running => true
     upgrade_step :remove_elasticsearch
+    upgrade_step :remove_docker_v1_content, :long_running => true
   end
 
   if [0,2].include? @kafo.exit_code


hooks/post/30-upgrade.rb
```diff
@@ -80,6 +87,7 @@ def fail_and_exit(message)
     upgrade_step :import_puppet_modules, :long_running => true
     upgrade_step :import_subscriptions, :long_running => true
     upgrade_step :remove_elasticsearch
+    upgrade_set :remove_docker_v1_content, :long_running => true
```
ehelms: s/set/step

bbuckingham: does this need to be done in hooks/pre/30-upgrade.rb to ensure that it happens prior to the rake db:migrate?

komidore64: @ehelms heh urite.

@bbuckingham i spoke with @ehelms and @jlsherrill over lunch and they seemed to think it would be fine before or after. If you look at the rake task, it's identifying the repositories through Runcible, rather than through the Katello model, then just telling the repository to destroy itself along with its clones.

bbuckingham: @komidore64, sounds good.  Thanks!  For consistency, I would prefer it in the post as well.   I assume that it will be tested before merged.  :)

komidore64: who tests things...?

komidore64:
re-running the travis build. don't think that error is related.

hooks/post/30-upgrade.rb
```diff
@@ -50,6 +50,13 @@ def remove_elasticsearch
   end
 end
 
+def remove_docker_v1_content
+  if Kafo::Helpers.execute('foreman-rake katello:upgrades:3.0:delete_docker_v1_content')
+    message = "Docker v1 repositories have been removed."
```
beav: This may be extraneous since the task itself has a message now.

beav:
If you decide to not print the message here, maybe just switch `remove_docker_v1_content` to just execute the task, sans if statement.

komidore64:
@beav ah yes. you are correct.

beav:
:heart_eyes: 

ACK

ehelms:
Thanks @komidore64 !

komidore64:
\o/

PR #296, opened 2016-02-12
Fixes #13687 - use the server cert for the certifiate_chain_file

iNecas:
the SSLCertificateChainFile is sent to the browser to verify the identity of
the server. The misconfiguration caused, that the user would need to have all
the certificates in the chain installed in the system. After this change, the
ROOT CA is enough to have on the system to verify the server cert.

### DIFF
--- a/config/katello.migrations/160802130342-certificate-chain-file.rb
+++ b/config/katello.migrations/160802130342-certificate-chain-file.rb
@@ -1,0 +1,3 @@
+if answers['foreman']['server_ssl_chain'] == '/etc/pki/katello/certs/katello-default-ca.crt'
+  answers['foreman']['server_ssl_chain'] = '/etc/pki/katello/certs/katello-server-ca.crt'
+end


beav:
I don't understand this change 100%. Is the expectation that `/etc/pki/katello/certs/katello-server-ca.crt` contains the full chain, and `/etc/pki/katello/certs/katello-default-ca.crt` does not?

iNecas:
The problem is the `SSLCertificateChainFile` is expected to contain the server ca certificates (for the https), while we're sending there the default ca certs (which are used for the client certs)

ehelms:
@iNecas looks like this needs rebasing and most likely a migration? Does this also solve http://projects.theforeman.org/issues/12265 ?

iNecas:
Seems like it might be related

ehelms:
Great - I'll merge once you rebase and this is passing.

ehelms:
@iNecas Can you please rebase?

iNecas:
Rebased and turned into migration

ehelms:
ACK - thanks @iNecas !

PR #297, opened 2016-02-22
Fixes #13250 - Remove capsule 9090 url from install

chris1984:
Removed Capsule line since url is not for customers to visit.


stbenjam:
FWIW, you can just amend your commit and force push to the branch instead of opening a new PR.  https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

PR #298, opened 2016-02-22
Fixes #13250 - Remove capsule 9090 url

chris1984:
removes url that is not for a customer to visit


ehelms:
ACK - Thanks @chris1984 !

PR #301, opened 2016-02-24
Refs #7064 - run puppet with --profile on Ruby 1.8.7

mbacovsky:
Re-add puppet settings lost during scenarification


ehelms:
ACK - thanks @mbacovsky !

PR #305, opened 2016-02-27
Fixes #13924: Ensure proper ordering for devel dependencies

ehelms:
Crane configuration ended up with the wrong certs and key because
certs::apache in the capsule module was not resolving when the
module was evaluated.

### DIFF
--- a/config/katello-devel.migrations/20160301144353-order.rb
+++ b/config/katello-devel.migrations/20160301144353-order.rb
@@ -1,0 +1,7 @@
+scenario[:order] = [
+  "certs",
+  "katello_devel",
+  "foreman_proxy",
+  "foreman_proxy::plugin::pulp",
+  "capsule"
+]


ehelms:
To test this, run an installation and check `/etc/httpd/conf.d/03-crane.conf` and ensure that it does not have "localhost.cert" set in the SSL directives.

mbacovsky:
With my scratch build it didn't worked. The installation completed fine, but the cert is still localhost.
It seems the reason is that on green field install the module order in scenario is updated (replaced actualy) by the migration. And it also seems that the migration is missing `katello-devel` and has `katello` instead. Adding `katello-devel` and re-running the installer fixed the certs.

I'd suggest to fix the migration if it was not released yet or add mew migration removing `katello` and adding  `katello-devel` in the order otherwise.

ehelms:
Updated to use a migration for those with existing devel installs

stbenjam:
Tested on my dev box and it corrected the issue, ACK from me

PR #306, opened 2016-02-29
Fixes #13951 - Add smart proxy plugin to katello scenario

ares:


ehelms:
Does this also  need to be included in the capsule scenario? https://github.com/Katello/katello-installer/blob/master/config/capsule-answers.yaml

ares:
probably yes, added

config/capsule-answers.yaml
```diff
@@ -3,3 +3,4 @@ certs:
   generate: false
   deploy: true
 capsule: true
+"foreman_proxy::plugin::remote_execution::ssh": false
```
ehelms: Sorry to nitpick, just not sure if it makes a difference: here you have it in quotes, while in katello-answers you don't.

ares:
removed quotes :-)

ehelms:
ACK - thanks @ares !

PR #307, opened 2016-02-29
fixes #13956: fix the installer rpm name needed to install a capsule

bbuckingham:
The 'capsule' scenario is included as part of the 'foreman-installer-katello' rpm.


johnpmitsch:
This is the package I needed to install the capsule, ACK from me

PR #309, opened 2016-03-03
Fixes #13198 - remove node importers and distributors from repos

johnpmitsch:
This removes node importers from repositories on a capsule and node distributors from repositories on the main katello server as they are no longer needed because of the [nodes removal work that has been done.](https://github.com/Katello/katello/pull/5778)

### DIFF
--- a/hooks/pre/30-upgrade.rb
+++ b/hooks/pre/30-upgrade.rb
@@ -40,6 +40,14 @@ def migrate_foreman
   Kafo::Helpers.execute('foreman-rake db:migrate')
 end
 
+def remove_nodes_importers
+  Kafo::Helpers.execute("mongo pulp_database --eval 'db.repo_importers.remove({'importer_type_id': \"nodes_http_importer\"});'")
+end
+
+def remove_nodes_distributors
+  Kafo::Helpers.execute("mongo pulp_database --eval  'db.repo_distributors.remove({'distributor_type_id': \"nodes_http_distributor\"});'")
+end
+
 def upgrade_step(step)
   noop = app_value(:noop) ? ' (noop)' : ''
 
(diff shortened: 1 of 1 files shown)


hooks/pre/30-upgrade.rb
```diff
@@ -60,15 +68,20 @@ def fail_and_exit(message)
...
   upgrade_step :start_databases
 
-  if Kafo::Helpers.module_enabled?(@kafo, 'katello') || @kafo.param('capsule', 'pulp').value
+  if Kafo::Helpers.module_enabled?(@kafo, 'katello') || @kafo.param('foreman_proxy_plugin_pulp', 'pulpnode_enabled').value
     upgrade_step :migrate_pulp
     upgrade_step :start_httpd
   end
 
+  if @kafo.param('foreman_proxy_plugin_pulp', 'pulpnode_enabled').value
+    upgrade_step :remove_nodes_importers
```
stbenjam: To make what's happening clearer, I might move this to the block above, remove the katello condition, and also add migrate_pulp, start_httpd in the katello if block below.

It's slightly less DRY but more readable. One can look and easily see "All these happen on the capsule", and "All those happen on the katello".

hooks/pre/30-upgrade.rb
```diff
@@ -40,6 +40,14 @@ def migrate_foreman
   Kafo::Helpers.execute('foreman-rake db:migrate')
 end
 
+def remove_nodes_importers
+  Kafo::Helpers.execute("mongo pulp_database --eval 'db.repo_importers.remove({'importer_type_id': \"nodes_http_importer\"});'")
+end
+
+def remove_nodes_distributors
+  Kafo::Helpers.execute("mongo pulp_database --eval  'db.repo_distributors.remove({'distributor_type_id': \"nodes_http_distributor\"});'")
```
ehelms: There is no way to do this through a Rake task using Runcible to update repositories within the Katello repository?

hooks/pre/30-upgrade.rb
johnpmitsch: @stbenjam I would think migrate_pulp and start_httpd should happen  on both a capsule and katello. If I move those to a block that only checks for katello, then they won't happen on a capsule as well?

This is how I read it now:
```
if capsule or katello
  migrate pulp
  start httpd

if capsule
  remove nodes importers

if katello
 migrate candlepin
 migrate foreman
 migrate gutterball
 remove nodes distributors

Comment (stbenjam) [inline hooks/pre/30-upgrade.rb]: I know, I'm suggesting to duplicate migrate_pulp and start_httpd so you only have 2 if statements. I'm ok if you want to leave it the way it is, I was just thinking it might be a little more readable.


```
if capsule
  migrate pulp
  start httpd
  remove nodes importers 

if katello
 migrate pulp
 start httpd
 migrate candlepin
 migrate foreman
 migrate gutterball
 remove nodes distributors
```

hooks/pre/30-upgrade.rb
jlsherrill: There is, but we need this to run on the capsule and the katello server, using mongo seemed like the easiest bet.

johnpmitsch:
updated to use variables for the katello/capsule checks to make things a little more clear. I have to completely test this out on a 2.4 instance and capsule install and will report back if it runs without issue.
.

johnpmitsch:
I tested this and the nodes removal part runs fine on both Katello and a capsule. I checked that the nodes importers were removed on the capsule and they were. I have yet to check the nodes distributors were removed from the main katello server because a [pulp bug](https://pulp.plan.io/issues/1753) but I would be surprised if they weren't.

johnpmitsch:
update: I can confirm this removes node distributors from repos on the katello server

jlsherrill:
ACK from me

johnpmitsch:
@ehelms :+1: or :-1:?

ehelms:
ACK - thanks @johnpmitsch

PR #311, opened 2016-03-08
Fixes #13809 - http restart error on upgrade

johnpmitsch:


ehelms:
From my testing and thinking of a better solution, I could not arrive at one. I am curious though if @ekohl or @stbenjam could think of any better ideas. If a more detailed recap of the issue is needed I can also provide.

stbenjam:
What created /etc/httpd/conf.d/pulp_content.conf?

jlsherrill:
@stbenjam updating to the new  pulp-server rpm lays this down

stbenjam:
I see.  So conf.modules.d seems like a default CentOS 7 thing, but we must've removed the default configuration previously?

I'm not sure how else to fix it, because we need apache up before kafo can run to correct the change (and puppet would correct it, right?)

One note, the change does need to make sure the line isn't already there, otherwise we'll keep adding it over and over again if the user tries --upgrade multiple times.

ehelms:
When 2.4 was released, the apache module did not config files in the proper place for EL7 but was fixed after the fact (https://github.com/puppetlabs/puppetlabs-apache/commit/e825422b0080e66b32cd05ed51b5ccae69325c74) so the httpd.conf that is deployed on 2.4 does not point to it.

stbenjam:
Okie dokie, sounds good to me, other than my note above. :+1:

johnpmitsch:
changed to check for the line before updating. I tested this out on a katello upgrade and it is working fine. I haven't had time to set up 2.4 capsule and upgrade to test, but I do not remember ever having httpd errors when running a 2.4 to 3.0 upgrade. I plan to still test to be %100 sure.

ehelms:
@stbenjam please re-review the change you requested

stbenjam:
>  @stbenjam please re-review the change you requested

Looks good to me.

PR #312, opened 2016-03-09
Fixes #14128 - Do not remove elasticsearch on upgrade

jlsherrill:


hooks/post/30-upgrade.rb
```diff
@@ -38,15 +38,13 @@ def import_subscriptions
...
 
   gems = ['ruby193-rubygem-tire', 'tfm-rubygem-tire', 'elasticsearch', 'sigar', 'snappy-java', 'lucene4-contrib', 'lucene4']
-  gems.each do |gem|
-    Kafo::Helpers.execute("rpm -e #{gem}")
-  end
-  message = "Elasticsearch has been removed as a dependency, the database files can be "\
-            "removed manually with #rm -rf /var/lib/elasticsearch"
+  message = "Elasticsearch has been removed as a dependency.  The database files can be "\
+            "removed manually with #rm -rf /var/lib/elasticsearch.  "
+  message += "Some packages are no longer needed and can be removed:  #rpm -e #{rpms.join(' ')}"
```
ehelms: Is it worth doing a check for the packages and /var/lib/elasticsearch and if none of them are found, suppress the message or parts therein?

jlsherrill: I don't know that its worth it.  The simple check at the top i think is good enough.

jlsherrill: To clarify, i don't think the added complexity is worth the user experience.

ehelms:
ACK

PR #315, opened 2016-03-17
fixes #14254 - only show admin password on first successful install

stbenjam:
Currently, the admin password continues to be shown on subsequent runs,
even if the user has already changed it. This creates a .installed file
in the Kafo config file directory after the first successful install, to
ensure we only print out the credentials on the very first run.


stbenjam:
Thanks @ehelms

PR #316, opened 2016-03-21
Fixes #14241 - removing old upgrade steps we no longer support

mccun934:
The follow-up PR for Katello itself is:

https://github.com/Katello/katello/pull/5906


ehelms:
The two issues marked here look like duplicates. Can we close one as a duplicate and then just set this PR to Fixes the remaining one?

mccun934:
@ehelms done

jlsherrill:
ACK

PR #317, opened 2016-03-23
Fixes #14327 - add openscap plugin support

ares:


shlomizadok:
@ares++

config/katello-answers.yaml
```diff
@@ -34,11 +34,13 @@ foreman::plugin::hooks: false
 foreman::plugin::tasks: true
 foreman::plugin::chef: false
 foreman::plugin::default_hostgroup: false
+foreman::plugin::openscap: false
 foreman::plugin::puppetdb: false
 foreman::plugin::setup: false
 foreman::plugin::templates: false
 foreman::plugin::remote_execution: false
 
+foreman_proxy::plugin::openscap: false
```
ehelms: I think the answers file shouldn't be altered directly, the migration handles that as far as I know -- @mbacovsky ?

ares: That's true, but it would make resulting answer file very unreadable after we have >15 migrations.. migrations are there to support existing setups before the change. I think that keeping answer file up to date with migrations is beneficial. I know that migration will override the value from answer file but it's more clear for anyone looking at repo or answer file on file system before first installer run.

ares:
Anything I could do to help with reviewing?

ehelms:
Just slipped past me -- looks good, thanks @ares !

PR #319, opened 2016-03-28
Fixes #13546 - Update tftp to true

chris1984:


stbenjam:
I would guess it should probably go to a capsule migration too then, but it's still up in the air what we enable by default.

One question I've had about these scenarios, how is this working in an upgrade case? It sounds like we'll be flipping TFTP on for users who explicitly turned it off in previous versions of Katello, right? If it can't be avoided in the move to the new installer, it needs to be prominently documented.

ehelms:
@stbenjam can you re-review this?

config/capsule.migrations/160425201608-tftp-true.rb
```diff
@@ -0,0 +1,5 @@
+# Change value to true
+
+answers['foreman_proxy'] = {
+  'tftp' => yes,
```
ehelms: I think this should just be true.

config/katello.migrations/160328095934-tftp-true.rb
```diff
@@ -0,0 +1,5 @@
+# Change value to true
+
+answers['foreman_proxy'] = {
+  'tftp' => yes,
```
ehelms: I think this should just be true.

stbenjam:
Not sure we actually need this now that I think about it. It is on by default in foreman proxy: https://github.com/theforeman/puppet-foreman_proxy/blob/master/manifests/params.pp#L186

chris1984:
@stbenjam care if i close this PR then?

stbenjam:
Nope, go ahead! Thanks @chris1984

PR #320, opened 2016-03-31
fixes #14416 - Add katello-installer/capsule-installer to inform user of new installer

bbuckingham:
This commit contains 2 small changes to help aide users in moving to the
new installer.  Rather than remove the old commands, we will include them;
however, simply echo the new command and exit with an error code.


ehelms:
ACK

bbuckingham:
Thanks @ehelms !

PR #324, opened 2016-04-18
Fixes #14675: Start tomcat on upgrade before Foreman migrations

ehelms:


hooks/pre/30-upgrade.rb
```diff
@@ -19,6 +19,10 @@ def migrate_candlepin
   Kafo::Helpers.execute("/usr/share/candlepin/cpdb --update --password #{Kafo::Helpers.read_cache_data('candlepin_db_password')}")
 end
 
+def start_tomcat
+  Kafo::Helpers.execute('katello-service start --only tomcat')
```
stbenjam: On el6, the service is called tomcat6 :-(

stbenjam: But the script ignores non-existent services, so you can say `--only tomcat,tomcat6`

ehelms:
@stbenjam Updated based on your suggestion

stbenjam:
ACK

PR #326, opened 2016-04-18
Fixes #14694: Enable oauth for pre-upgrade database migrations

ehelms:


ekohl:
I don't know pulp well enough, but is it possible that oauth was on before the migration and this in effect turns it off?

ehelms:
For 3.0, we want it turned off, as we are moving to certificate based authentication for communication with the main Pulp and remote Pulps. However, we kept it in place in case we needed it still since communication with certificates is new. @jlsherrill or @johnpmitsch can elaborate or correct me if I have strayed

johnpmitsch:
@ehelms that is correct, we have switched to cert authentication for the default communication between pulp servers, but oauth is still available as a back up.

jlsherrill:
ACK

PR #328, opened 2016-04-25
Fixes #14750 - Proxy does not run on http by default

xprazak2:


ekohl:
Should this be a new migration instead of changing the existing one?

xprazak2:
I created a new migration, not changing the existing one anymore.

stbenjam:
Whoa! :confused: 

There's a very good reason HTTP is **on**.  This is definitely a no-go for me, at least for the capsule part, unless we can figure it out to turn it on only when it is needed.  It's somewhat complicated.

Templates feature uses HTTP for kickstarts as Anaconda needs it over HTTP. It's required to isolate communication from a host to the capsule.   Templates is (or should be) on by default on a capsule, so HTTP needs to be enabled as well.  If the user wants to turn it off then they need to do `--foreman-proxy-templates=false --foreman-proxy-http=false`

Having HTTP on is not exposing any private services over HTTP - the default for the installer is that features like DHCP, DNS, etc ONLY listen on HTTPS unless a user explicitly says `--foreman-proxy-dhcp-listen-on=both` in the installer. One exception is Templates, which listens on both by default.

So, I'm ok with the Katello change, I can't think of a reason it needs to listen on HTTP, but for capsule we need to leave it on.

stbenjam:
One more thought, I've mentioned before but we really need to get resolution on it.

These migrations we keep adding are also kind of broken behavior if we're running them on a Katello installation in an upgrade -- particularly a katello that predated the scenario work.

If a user has explicitly turned http on, you're turning it off for them in a migration.  We also have another PR where we turn TFTP on, but a user may have explicitly turned that off before.  We need some clarification on this from an upgrades perspective. How are 2.x -> 3.0 upgrades affected?

Do we just not run these migrations on pre-scenario installs? @ehelms  @bbuckingham @mbacovsky

xprazak2:
Oh, I did not know about Anaconda and that Templates are on by default for `scenario --capsule`, I will leave it alone, turn it off for katello and wait for suggestions on migrations.

xprazak2:
@stbenjam, seems like overriding user changes is not a pressing issue and it is out of scope for this PR anyway. Would it be possible to move the discussion to mailing list and conclude this?

config/katello.migrations/160425017684-default-http.rb
```diff
@@ -0,0 +1,3 @@
+answers['foreman_proxy']['http'] = false if answers['foreman_proxy'].is_a? Hash
+
+answers['foreman_proxy'][('http' || 'foreman_proxy_http')] = answers['capsule'].delete('http') { |k| false } if answers['capsule'].is_a? Hash
```
stbenjam: What is the `'http' || 'foreman_proxy_http'` supposed to be doing? That will always resolve to http as non-blank strings are truthy.

stbenjam:
> @stbenjam, seems like overriding user changes is not a pressing issue and it is out of scope for this PR anyway. Would it be possible to move the discussion to mailing list and conclude this?


That seems fine, looks good now other than the one question.

xprazak2:
Right, should be ok now.

config/katello.migrations/160425017684-default-http.rb
```diff
@@ -0,0 +1,3 @@
+answers['foreman_proxy']['http'] = false if answers['foreman_proxy'].is_a? Hash
+
+answers['capsule'].delete('foreman_proxy_http') if answers['capsule'].is_a? Hash
```
stbenjam: I'm still confused why this is here, before it was setting the value in foreman_proxy from the old capsule value (which was already done: https://github.com/Katello/katello-installer/blob/master/config/katello.migrations/160120212701-foreman-proxy.rb#L67), but now its just deleting it.  That's not required either, Kafo will do this since the param doesn't exist anymore.

xprazak2:
You are right, it is not needed.

xprazak2:
Ah, wrong button.

stbenjam:
Looks great, thanks! ACK

ekohl:
Thanks!

PR #329, opened 2016-04-27
Refs #14858 - removes gutterball

cfouant:
Main PR: https://github.com/Katello/katello-packaging/pull/213

### DIFF
--- a/config/katello-answers.yaml
+++ b/config/katello-answers.yaml
@@ -46,8 +46,6 @@ foreman::plugin::remote_execution: false
 foreman_proxy::plugin::openscap: false
 foreman_proxy::plugin::remote_execution::ssh: false
 
-katello::plugin::gutterball: true
-
 # Issue #4244 - Commenting these out due to the seg fault
 # caused on Ruby 1.8.7, the Katello RPM ensures these
 # are installed instead
--- a/config/katello-devel.migrations/160503102719-gutterball.rb
+++ b/config/katello-devel.migrations/160503102719-gutterball.rb
@@ -1,0 +1,2 @@
+answers.delete('katello_devel::plugin::gutterball') if answers.include?('katello_devel::plugin::gutterball')
+answers.delete('katello_devel::plugin::foreman_gutterball') if answers.include?('katello_devel::plugin::foreman_gutterball')
--- a/config/katello.migrations/160503102427-gutterball.rb
+++ b/config/katello.migrations/160503102427-gutterball.rb
@@ -1,0 +1,1 @@
+answers.delete('katello::plugin::gutterball') if answers.include?('katello::plugin::gutterball')
--- a/hooks/post/30-upgrade.rb
+++ b/hooks/post/30-upgrade.rb
@@ -40,6 +40,11 @@ def remove_docker_v1_content
   Kafo::Helpers.execute('foreman-rake katello:upgrades:3.0:delete_docker_v1_content')
 end
 
+def remove_gutterball
+  return true unless Kafo::Helpers.execute('rpm -q gutterball')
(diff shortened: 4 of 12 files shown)


stbenjam:
Needs rebase

hooks/post/30-upgrade.rb
```diff
@@ -54,6 +54,11 @@ def remove_docker_v1_content
   Kafo::Helpers.execute('foreman-rake katello:upgrades:3.0:delete_docker_v1_content')
 end
 
+def remove_gutterball
+  return true unless Kafo::Helpers.execute('rpm -q gutterball')
+  Kafo::Helpers.execute("rpm -e gutterball")
```
stbenjam: I think we also need:

- tfm-rubygem-hammer_cli_gutterball
- gutterball-certs
- gutterball
- tfm-rubygem-foreman_gutterball

If users are coming from older katello (and Satellite), there's another wrinkle those packages could be ruby193- and not tfm

config/katello-answers.yaml
```diff
@@ -41,8 +41,6 @@ foreman::plugin::remote_execution: false
 
 foreman_proxy::plugin::remote_execution::ssh: false
 
-katello::plugin::gutterball: true
```
stbenjam: Existing installs won't get this update, so we'll need a migration to remove this from existing answer files too, see `config/migrations/` dir for some examples

stbenjam:
I didn't try all the PR's together but at first glance it looks good :+1: other than the couple of comments.  If you rebase these two PR's I can try them all together and see how it goes.

hooks/post/30-upgrade.rb
ehelms: Although this happens in post- and thus all the packages should have already been upgraded to tfm-

ehelms:
Would you mind providing a list of all the PRs related to this on whichever is the "major" PR for easier tracking? Also, updating all of the non-major PRs to use 'Refs' instead of fixes. That way, if one of those gets merged, the issue does not auto-close making us think its complete and leaving a half-cherry picked change for 3.0.

cfouant:
@ehelms @stbenjam thanks for your initial comments, i've made all edits

config/katello.migrations/160503102427-gutterball.rb
```diff
@@ -0,0 +1 @@
+answers['katello::plugin::gutterball'].delete if answers['katello::plugin::gutterball']
```
stbenjam: Was trying this out and hit this error:
`/etc/foreman-installer/scenarios.d/katello.migrations/160503102427-gutterball.rb:1:in `delete': wrong number of arguments (0 for 1) (ArgumentError)`

I think it needs to be `answers.delete("katello::plugin::gutterball")`  (and same on the other migrations in this pr)

cfouant:
@stbenjam thanks, changes made

config/katello-devel.migrations/160503102719-gutterball.rb
```diff
@@ -0,0 +1,2 @@
+answers.delete('katello_devel::plugin::gutterball') if answers['katello_devel::plugin::gutterball']
+answers.delete('katello_devel::plugin::foreman_gutterball') if answers['katello_devel::plugin::foreman_gutterball']
```
ehelms: If for some reason these were set to false, I don't think this would actually remove them.

config/katello.migrations/160503102427-gutterball.rb
```diff
@@ -0,0 +1 @@
+answers.delete('katello::plugin::gutterball') if answers['katello::plugin::gutterball']
```
ehelms: If this was ever set to false, then it wouldn't get removed.

stbenjam:
ACK

stbenjam:
ACK

PR #331, opened 2016-05-09
Fixes #14956 - Check for proxy before katello-installer runs

chris1984:


checks/proxy.rb
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env ruby
+
+begin
+  require 'rubygems'
+rescue LoadError
+end
```
stbenjam: This isn't needed

checks/proxy.rb
```diff
@@ -0,0 +1,20 @@
...
+
+# checking to see if http/https proxy envs are set before running the installer
+
+http_proxy = ENV['http_proxy']
+https_proxy = ENV['https_proxy']
+
+PROXYSET = "Currently there is a environmental proxy set to reach out to the internet, Katello does a check via wget to make sure services are running correctly. Please unset this and then run katello-installer again."
```
stbenjam: Nitpick, but this shouldn't be a constant (all uppercase creates a constant), it'll never be garbage collected

stbenjam:
What about just adding `--no-proxy` to any of our calls that connect to the localhost?

checks/proxy.rb
stbenjam: Nevermind. I guess the other checks do this, so seems to be a thing we do.

chris1984:
@stbenjam Thanks for the help, I removed the gem section and fixed some of the wording to help with what you stated.

chris1984:
@stbenjam I can add a 2nd PR/Issue which would rule out this one but my concern is if we ever add anything to service-wait we would want to enforce all future wget commands with --no-proxy where this would be a catch all. Being new to putting patches in I think we should have this and a 2nd issue for the --no-proxy but I may be offbase here.

stbenjam:
Code's good :+1: I think you are right that `--no-proxy` everywhere might be a lot and we'd be opening ourselves to future problems if we miss one.  I'm just wondering if there's any other implications here, I'm hoping someone else from @Katello/installer will chime in :)

ekohl:
Are there valid use cases where people would want a proxy during the installer? I'm mostly thinking about yum repos, but I suppose they could configure yum to use a proxy.

chris1984:
@ekohl I have seen quite a few use cases downstream for things such as curl/wget and a few other applications. I know it has caused some issues especially if users forget its set.

ekohl:
@chris1984 I meant use cases where the users really wants a http proxy set while running katello-installer where this would break their use case. I can only think of yum and users can work around that through yum.conf.

stbenjam:
@ekohl Indeed, that's what I was trying to think of - yum can have it's own proxy set, I can't think of anything else we might need it for.

ekohl:
ACK I think we aim to not need any network access besides (RPM) packages and should consider any other required network access a bug.

chris1984:
@ekohl I agree with your last comment that a proxy should not be set during the install. yum has its own proxy settings in yum.conf and rhsm has its own proxy in rhsm.conf. I think this is a valid check.

stbenjam:
Thanks @chris1984, this works great:

```
[root@katelloprod checks]# http_proxy=www.google.com foreman-installer
Currently there is a environmental proxy variable set. Katello does a check on services when running the installer to make sure services are running correctly. Please unset this and then run katello-installer again. Once the installer finishes you can set the proxy variables back.
```

One little nitpick though, could we make the test more helpful? An inexperienced user may not understand what environmental proxy means

I'd prefer if we were very explicit.  Something like "Please unset the http_proxy and/or https_proxy environment variables before running the installer"

chris1984:
Closing this PR because of me deleting the forked repo and a new fork with same branch not pushing to this PR correctly. New PR is https://github.com/Katello/katello-installer/pull/353

PR #332, opened 2016-05-10
Fixes #14978 - capsule-certs-generate does not honour "--parent-fqdn" parameter

ahumbe:
capsule-certs-generate does not honour "--parent-fqdn" parameter 

The parent fqdn is set as the actual hostname of the satellite server where we execute capsule-certs-generate command.

### DIFF
--- a/hooks/post/10-post_install.rb
+++ b/hooks/post/10-post_install.rb
@@ -17,7 +17,7 @@ end
 
 if [0,2].include?(@kafo.exit_code)
   if !app_value(:upgrade)
-    fqdn = Facter.value(:fqdn)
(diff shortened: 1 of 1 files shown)


ehelms:
Thanks @ahumbe !

hooks/post/10-post_install.rb
```diff
@@ -17,7 +17,7 @@ def new_install?
 
 if [0,2].include?(@kafo.exit_code)
   if !app_value(:upgrade)
-    fqdn = Facter.value(:fqdn)
+    fqdn = @kafo.param('capsule_certs','parent_fqdn').value || Facter.value(:fqdn)
```
akofink: This introduces a NoMethodError when calling `value` on `nil`.

PR #333, opened 2016-05-11
fixes #15014 - don't start mongo in reset hook

stbenjam:
Leaving mongo stopped will trigger database migration.  Currently reset
is failing because the database is never migrated on the reset case.

Requires https://github.com/Katello/puppet-pulp/pull/145


ehelms:
ACK

stbenjam:
This isn't needed as I'll just restore the old init.flag

PR #334, opened 2016-05-16
fixes #15063 - Set answers for foreman client certs

ekohl:
https://github.com/theforeman/puppet-foreman/commit/ec8c6c25356429b01cd590158932664271db387e exposes the foreman client certificates through parameters. Since Katello wants to change these from puppet to custom handling, these must be set. puppet-certs currently attempts to do so through foreman-rake config. Since puppet-foreman has started to set these in settings.yaml, they can't be overridden anymore. By setting these in the answers file, we avoid that conflict.

Note I didn't test this yet, but I believe this will become a problem so I opened this to track it.

### DIFF
--- a/config/answers.sam-installer.yaml
+++ b/config/answers.sam-installer.yaml
@@ -19,6 +19,9 @@ foreman:
   server_ssl_key: /etc/pki/katello/private/katello-apache.key
   server_ssl_ca: /etc/pki/katello/certs/katello-default-ca.crt
   server_ssl_chain: /etc/pki/katello/certs/katello-default-ca.crt
+  client_ssl_cert: /etc/pki/katello/certs/katello-apache.crt
+  client_ssl_key: /etc/pki/katello/private/katello-apache.key
+  client_ssl_ca: /etc/pki/katello/certs/katello-default-ca.crt
   websockets_encrypt: true
   websockets_ssl_key: /etc/pki/katello/private/katello-apache.key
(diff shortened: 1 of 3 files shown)


stbenjam:
This replaces https://github.com/Katello/puppet-certs/pull/87, right?

stbenjam:
Need to get this in ASAP, as it seems to have broken our nightly install:

```
 /Stage[main]/Certs::Foreman/Exec[foreman_certs_config]/returns: change from notrun to 0 failed: /usr/share/foreman/script/foreman-rake config --      -k ssl_ca_file -v '/etc/foreman/proxy_ca.pem'      -k ssl_certificate -v '/etc/foreman/client_cert.pem'      -k ssl_priv_key -v '/etc/foreman/client_key.pem' returned 1 instead of one of [0]
```

stbenjam:
I think the paths should be:

```
  $foreman_client_cert    = '/etc/foreman/client_cert.pem'
  $foreman_client_key     = '/etc/foreman/client_key.pem'
  $foreman_ssl_ca_cert    = '/etc/foreman/proxy_ca.pem'
```

stbenjam:
With these paths, and removing the stanzas from puppet-certs (the ones https://github.com/Katello/puppet-certs/pull/87 tried to change) entirely, I get a clean install.

stbenjam:
Closed in favor of https://github.com/Katello/katello-installer/pull/335

PR #335, opened 2016-05-16
fixes #15063 - set answers for foreman client certs

stbenjam:
puppet-foreman now exposes the foreman client certificates through
parameters. Since Katello wants to change these from puppet to custom
handling, these must be set. puppet-certs currently attempts to do so
through foreman-rake config. Since puppet-foreman has started to set
these in settings.yaml, they can't be overridden anymore. By setting
these in the answers file, we avoid that conflict.


stbenjam:
Needs https://github.com/Katello/puppet-certs/pull/88, too

beav:
can `answers.sam-installer.yaml` go away? I don't think it is used anywhere.

stbenjam:
Not sure, @thomasmckay would know.  If it can, I'd say it should be removed in a different issue though, one that specifically tracks sam removal.

beav:
```
  Success!
```

ACK

stbenjam:
Awesome!

Thanks @ekohl!

PR #336, opened 2016-05-17
Fixes #15072 - RHEL 6 http conf doesn't have conf.d dir structure

johnpmitsch:
RHEL 6 uses http 2.2 which does not have the conf.d/ directory structure. adding 
```Include "/etc/httpd/conf.modules.d/*.conf"``` causes errors in RHEL 6.

This adds a check for http 2.4 in the http.conf, which will only include the conf.d/ directory in RHEL 7 systems

### DIFF
--- a/hooks/pre/30-upgrade.rb
+++ b/hooks/pre/30-upgrade.rb
@@ -11,8 +11,9 @@ def start_httpd
 end
 
 def update_http_conf
-  Kafo::Helpers.execute("grep -q -F 'Include \"/etc/httpd/conf.modules.d/*.conf\"' /etc/httpd/conf/httpd.conf || \
-		         echo 'Include \"/etc/httpd/conf.modules.d/*.conf\"' >> /etc/httpd/conf/httpd.conf")
+  Kafo::Helpers.execute("grep -F -q 'Include \"/etc/httpd/conf.modules.d/*.conf\"' /etc/httpd/conf/httpd.conf || \
+                       echo -e '<IfVersion >= 2.4> \n    Include \"/etc/httpd/conf.modules.d/*.conf\"\n</IfVersion>' \
+                       >> /etc/httpd/conf/httpd.conf")
 end
 
 def migrate_candlepin


ekohl:
Why is that line even needed? Doesn't puppetlabs-apache fix that for us or is this for the non-vhost case where puppetlabs-apache isn't used?

johnpmitsch:
@ekohl it may not be needed, discussing it with @ehelms now

ehelms:
Is the question do we need this upgrade step? or a specific aspect of this upgrade step?

ekohl:
I wonder why the step ```update_http_conf``` is needed at all. It sounds like something that should be done in puppet.

ehelms:
During an upgrade, Pulp lays down `pulp_content.conf` which includes a reference to an mod_xsendfile directive. The mod_xsendfile RPM installs itself into the right path on the system, however, apache from 2.4 was configured to look at `/etc/httpd/conf.d` and not at `/etc/httpd/conf.modules/d` which was corrected in a later version of puppetlabs-apache. Thus, when we restart httpd it fails because it can't find where the mod_xsendfile directive comes from.

johnpmitsch:
I'll test removing this function and upgrade step on an upgrade and see if it is still needed

johnpmitsch:
@ehelms @ekohl so I tested out removing this function completely and it resulted in the mod_xsendfile error that @ehelms mentioned above. The installer ran without error with this change, but the /etc/httpd/conf/httpd.conf doesn't have the <IfVersion >= 2.4> part, which makes me think the installer is overwriting this change? I don't know if that is expected and this line is just so we have the ability to restart httpd before the installer runs.

I am spinning up a rhel 6 box to test further.

johnpmitsch:
I can't get a rhel 6 box up and running to test this. [here is the error if you are curious](https://paste.fedoraproject.org/368145/46359893/) I may do a manual install and test. Will update accordingly

ekohl:
Sounds like you were bitten by https://github.com/Katello/puppet-pulp/pull/147

johnpmitsch:
I got a EL6 box up and running and this change ran fine on a 2.4 > 3.0 upgrade

bbuckingham:
@johnpmitsch, I have tested this patch on both EL6 and EL7 upgrades and worked well on both.  ACK from me.

ehelms:
Thanks @johnpmitsch !

PR #337, opened 2016-05-19
Fixes #15062 - Check for active tasks before upgrade

daviddavis:
Depends on https://github.com/theforeman/foreman-tasks/pull/186

### DIFF
--- a/hooks/pre/30-upgrade.rb
+++ b/hooks/pre/30-upgrade.rb
@@ -1,3 +1,5 @@
+require 'tempfile'
+
 def stop_services
   Kafo::Helpers.execute('katello-service stop --exclude mongod,postgresql')
 end
@@ -74,11 +76,24 @@ def fail_and_exit(message)
   kafo.class.exit 1
 end
 
+def check_active_tasks
+  filepath = Tempfile.new("katello_upgrade_tasks").path
+  filter = "state != stopped && label != Actions::Candlepin::ListenOnCandlepinEvents"
+
+  Kafo::Helpers.execute("foreman-rake foreman_tasks:export_tasks TASK_SEARCH=\"#{filter}\" TASK_FORMAT=csv TASK_FILE=\"#{filepath}\"")
+  task_count = `wc -l "#{filepath}"`.strip.split.first.to_i - 1
+
+  if task_count > 0
+    fail_and_exit "Upgrade cannot proceed. You have #{task_count} active task(s)."
+  end
+end
+
 if app_value(:upgrade)
   Kafo::Helpers.log_and_say :info, 'Upgrading...'
   katello = Kafo::Helpers.module_enabled?(@kafo, 'katello')
   capsule = @kafo.param('foreman_proxy_plugin_pulp', 'pulpnode_enabled').value
 
+  check_active_tasks
   upgrade_step :stop_services
   upgrade_step :start_databases
   upgrade_step :update_http_conf if Kafo::Helpers.module_enabled?(@kafo, 'katello')


hooks/pre/30-upgrade.rb
```diff
@@ -74,11 +74,24 @@ def fail_and_exit(message)
   kafo.class.exit 1
 end
 
+def check_active_tasks
+  filename = "/tmp/katello_upgrade_tasks#{Time.now.to_i}"
```
ekohl: Why don't you use ```Tempfile``` for this?

hooks/pre/30-upgrade.rb
```diff
@@ -74,11 +74,24 @@ def fail_and_exit(message)
...
+  filter = "state != stopped && label != Actions::Candlepin::ListenOnCandlepinEvents"
+
+  Kafo::Helpers.execute("foreman-rake foreman_tasks:export_tasks FILTER=\"#{filter}\" FORMAT=csv FILE=\"#{filename}\"")
+  task_count = `wc -l "#{filename}.csv"`.strip.split.first.to_i - 1
```
iNecas: I've put a comment in https://github.com/theforeman/foreman-tasks/pull/186/files#r64035793 when seeing here how we need to handle the suffixes.

daviddavis:
After talking with some other folks, it sounds like this might be a bad idea. Basically this could leave an upgrade in a bad state:

1. User upgrades katello/candlepin/etc packages
2. User runs upgrade which fails due to active tasks
3. System is unusable as candlepin, etc has been upgraded and therefore the user cannot resume the tasks.

ekohl:
Another thing to consider is that pretty much want to put foreman-tasks in a hold state where it won't start new tasks and resume when it's done. Otherwise you check in a pre-hook but a task is created after running the hook.

iNecas:
Ideally, a pre-upgrade script that itself would run the update (including yum repositories)? Also, the check should happen before the candlepin/pulp dbs are upgreded, so that it would be save to downgrade the rpms and finish it first.

Yes @ekohl  the maintenence mode would be nice and that's something we should work in min-term horizont

ehelms:
We were tossing around the idea of a separate tool (foreup? tfm-up? foreman-up? foreman-upgrade?) that would handle all of the necessary upgrade checks and steps and leave the installer to do what it does. Or, we just bake this into the installer so it's an all-in-one handler of everything.

daviddavis:
@ehelms Just to be clear—you're saying that these tools would take care of everything from yum update to checking the system for active tasks? I'm ok with either of those solutions. Also, would this be for katello 3.0?

ehelms:
I think that is what we are thinking and would give us more flexibility and control over the process. We would need to have it in the prior release to be effective though so that packages are not upgraded. Potentially, we could have users update the repos and instruct them to install the upgrade package or installer only and let that handle the rest. Given the newness of such an approach, I'd think we would want to provide it experimentally for a release. That is, ensure there is a manual process users can follow to upgrade but also provide the new tooling. That may be too much work for 3.0 at this point though.

iNecas:
Having the tool not limited to katello only, but to foreman in general would be very welcome.

mccun934:
+1 to foreman and not katello specific

How about something like:

```
## While at Foreman 1.6 / Katello 2.4.0 or other older version of Foreman

# foreman-upgrade-check

1) Task Check:
[ERROR] You have 5 paused tasks that need to be resumed or killed before proceeding with upgrading

2) Version Check:
[ERROR] Your Foreman is at 1.6.3 and needs to be upgraded to 1.11.X before upgrading

3) Host Unification Report:
[OK] There are no Content Hosts that will be removed after upgrading

## While at 1.11:

# foreman-upgrade-check

1) Task Check:
[ERROR] You have 5 paused tasks that need to be resumed or killed before proceeding with upgrading

2) Version Check:
[OK] Your Foreman is at the correct version to upgrade to the latest version

3) Host Unification Report:
[OK] There are no Content Hosts that will be removed after upgrading to Katello 3.0

## All clear for 1.11

# foreman-upgrade-check

1) Task Check:
[OK] No paused/pending tasks

2) Version Check:
[OK] Your Foreman is at the correct version to upgrade

3) Host Unification Report:
[OK] There are no Content Hosts that will be removed after upgrading 
```

stbenjam:
Yea, it could take a debug approach and have something like `upgrade-check.d` where plugins can dump in their checks?  Is that overkill though? Not sure anyone else other than Katello *needs* this yet.  Foreman needs to address more fundamental upgrade issues.

> @inecas: Having the tool not limited to katello only, but to foreman in general would be very welcome.

Foreman basically leaves users to handle it all themselves, they don't use the installer to handle upgrades, although its kind of recommended. If users do it though, there's no guarantee things aren't going to get messed up.  Between foreman versions, puppet module parameters change, and there's no effort AFAIK to create migrations for answers.

ehelms:
At least using foreman naming conventions is half the battle :) The check style script is a good bridge towards a more formal full upgrade tool/process.

ekohl:
> Foreman basically leaves users to handle it all themselves, they don't use the installer to handle upgrades, although its kind of recommended. If users do it though, there's no guarantee things aren't going to get messed up. Between foreman versions, puppet module parameters change, and there's no effort AFAIK to create migrations for answers.

Recently we have started [adding answer migrations](https://github.com/theforeman/foreman-installer/tree/develop/config/foreman.migrations). I wonder what @domcleal and @mmoll think about concentrating these efforts.

Personally I'd like to see an easy and extensible solution that works for both foreman and katello.

stbenjam:
> Recently we have started adding answer migrations.

Oh awesome! I totally missed that.

mccun934:
the check script is definitely intended to be a short-term bridge, we can fold the actual checks into something more automated later.

need something for Katello 2.4 -> 3.0

ehelms:
The only other question for me would be the delivery mechanism:

 1) Deliver in 2.4 directly
 2) Deliver in 3.0 and update instructions to have users install and run script before hand

stbenjam:
How about both? Users should do it in 2.4, but clearly there will be people who just yum upgrade, and will still need to run the script.

mccun934:
+1 to both 2.4 and 3.0

PR #338, opened 2016-05-20
Fixes #15117 - Pre script to handle setting/unsetting of pulp oauth

parthaa:
One of the things that changed after katello 2.4 was the removal of
oauth. To enable a smooth transtition we need to change the upgrade
scripts to do the following

1) Set pulp to use Oauth
2) Complete the migration via -> https://github.com/theforeman/puppet-foreman/blob/master/manifests/database.pp#L29 .
3) Unset the use of pulp oauth after migration so that it uses the pulp cert auth before seeding

This commit tries to achieve that

### DIFF
--- a/hooks/post/30-upgrade.rb
+++ b/hooks/post/30-upgrade.rb
@@ -40,6 +40,10 @@ def remove_docker_v1_content
   Kafo::Helpers.execute('foreman-rake katello:upgrades:3.0:delete_docker_v1_content')
 end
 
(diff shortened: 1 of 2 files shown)


ehelms:
Katello upgrades from 2.4 to 3.0 work fine in all my scenarios I test with
the RCs. Are you seeing this happen on 2.3 migrations? Or have a reproducer?

ehelms:
Closed given #339

PR #339, opened 2016-05-21
Fixes #15117 - Upgrade steps required for 2.3 to 3.0

ehelms:
This is an alternative approach to #338 where I am centralizing all the "unique" logic required for these upgrades into the pre-upgrade script and attempting to provide as much idempotent behavior as possible. Another approach would be to add "upgrade.d" functionality similar to migrations and mark upgrades that have run and perform these there. 

Note: This is only from a 2.3 to 3.0 upgrade which we say we support currently. There is one outstanding issue with this that I did not include a fix for yet. The Foreman PR (https://github.com/theforeman/foreman/pull/3519) is not currently in 1.11 and the migrations fail because of it. As I see it our options then are:

 1) Push to get https://github.com/theforeman/foreman/pull/3519 into 1.11.3 and wait for GA until then
 2) Add upgrade instructions that do the following (and yes this worked testing locally):
   - disable katello as a gem
   - run foreman-rake db:migrate to migrate Foreman
   - renable katello as a gem
   - re-run foreman-rake db:migrate to migrate Katello
 2) Require users to upgrade to 2.4 and then to 3.0

### DIFF
--- a/hooks/pre/30-upgrade.rb
+++ b/hooks/pre/30-upgrade.rb
@@ -10,6 +10,14 @@ def start_httpd
   Kafo::Helpers.execute('katello-service start --only httpd')
 end
 
+def start_qpidd
+  Kafo::Helpers.execute('katello-service start --only qpidd,qdrouterd')
+end
+
+def start_pulp
+  Kafo::Helpers.execute('katello-service start --only pulp_workers,pulp_resource_manager,pulp_celerybeat')
+end
+
 def update_http_conf
   Kafo::Helpers.execute("grep -q -F 'Include \"/etc/httpd/conf.modules.d/*.conf\"' /etc/httpd/conf/httpd.conf || \
 		         echo 'Include \"/etc/httpd/conf.modules.d/*.conf\"' >> /etc/httpd/conf/httpd.conf")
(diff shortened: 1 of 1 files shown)


bbuckingham:
Tested an upgrade and the results were positive.  ACK from me; however, I'd like to get a second pair of eyes to look over it.

jlsherrill:
ACK from me given @bbuckingham's testing

ehelms:
Could you guys please weigh in on my question from above starting with the 'Note:' section?

hooks/pre/30-upgrade.rb
```diff
@@ -59,6 +67,24 @@ def remove_nodes_distributors
   Kafo::Helpers.execute("mongo pulp_database --eval  'db.repo_distributors.remove({'distributor_type_id': \"nodes_http_distributor\"});'")
 end
 
+def fix_pulp_httpd_conf
+  return true unless File.exist?('/etc/httpd/conf.d/pulp.conf.rpmnew')
+
+  Kafo::Helpers.execute('cp /etc/httpd/conf.d/pulp.conf.rpmnew /etc/httpd/conf.d/pulp.conf')
+end
```
ehelms: Does this seem idempotent enough given upgrades are run between z-releases as well?

jlsherrill:
I'd probably document that if you are upgrading from 2.3 run these extra steps.

hooks/pre/30-upgrade.rb
bbuckingham: It is hard to be 100% certain; however, given the types of changes that go in to the z-releases, it seems like this should still be Ok.  That said, we'll need to ensure that we continue to always run --upgrade on those releases and perform proper regression testing.

I did perform multiple --upgrade runs with the installer using these changes and have not seen any adverse affects; however, I do realize that those tests do not involve additional software updates being introduced.

hooks/pre/30-upgrade.rb
```diff
@@ -59,6 +67,24 @@ def remove_nodes_distributors
...
+
+def fix_katello_settings_file
+  settings_file = '/etc/foreman/plugins/katello.yaml'
+  settings = JSON.parse(JSON.dump(YAML.load_file(settings_file)), symbolize_names: true)
```
bbuckingham: Note: we'll need to update this to be compatible on EL6.

```
/usr/share/katello-installer-base/hooks/pre/30-upgrade.rb:73: syntax error, unexpected ':', expecting ')'
...tings_file)), symbolize_names: true)
```

ehelms:
Updated @bbuckingham

bbuckingham:
@ehelms, Thanks for the update.  I've tested both EL6 & EL7 and the behavior looks good. ACK for me.

PR #340, opened 2016-05-24
Fixes #15153 - Use actual installer name in post-install msg

mbacovsky:


ekohl:
Could you explain why this is sometimes foreman-installer and sometimes katello-installer?

mbacovsky:
@ekohl, before `Katello 3.0` Katello had its own installer called `katello-installer`. With 3.0 we introduced scenario based common installer `foreman-installer` for the whole project and `katello-installer` became scenario installable with `foreman-installer`. This repo name remained intact, but `katello-installer` as a installer script no longer exist. Currently we have only `foreman-installer` for installing everything in upstream. But in downstream the installer is called `satellite-installer` and that is why I'm proposing this change

ekohl:
That helps, but isn't simply using the invocation-path sufficient? IIRC that used the filename that called it, which would be satellite-installer for a downstream release.

mbacovsky:
@ekohl, that was my assumption as well, but `capsule-certs-generate` script is using the same hook and `invocation_path` would be set to `capsule-certs-generate` instead of `foreman-installer`

ekohl:
That makes sense, ACK

mbacovsky:
Thanks!, @ekohl

ehelms:
Thanks @mbacovsky !

PR #341, opened 2016-05-24
fixes #15158 - support multiple paths for katello-certs-check

stbenjam:
katello-certs-check is either in the katello-installer directory
(git repo) or in foreman-installer-katello (rpm install).  This supports
both cases.

For 3.1, would it just make sense to combine these into 2 packages? I wasn't sure if there's a technical reason for why we keep foreman-installer-katello and katello-installer-base as 2 separate packages.

### DIFF
--- a/hooks/pre/20-certs_update.rb
+++ b/hooks/pre/20-certs_update.rb
@@ -1,7 +1,7 @@
 require 'fileutils'
 
 SSL_BUILD_DIR = '/root/ssl-build/'
-CHECK_SCRIPT  = File.expand_path('../../../bin/katello-certs-check', __FILE__)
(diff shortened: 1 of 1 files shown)


ehelms:
I wonder if we should just install the script via RPM so its available system wide and avoid path wars?

stbenjam:
> I wonder if we should just install the script via RPM so its available system wide and avoid path wars?

Yea, it is in $PATH so we could just do that now, except for development installs.  Is it ever needed there?

ehelms:
For testing and using custom certs I could see it being useful.

stbenjam:
Ah, well, it is in the package even on dev installs, and it's in $PATH.  So, updated to use which.

ehelms:
ACK

PR #342, opened 2016-05-24
fixes #15167 - let user know they can specify org/locations

stbenjam:
This goes with https://github.com/theforeman/puppet-foreman_proxy/pull/253 https://github.com/theforeman/puppet-foreman/pull/448.

### DIFF
--- a/hooks/post/10-post_install.rb
+++ b/hooks/post/10-post_install.rb
@@ -44,23 +44,22 @@ MSG
         foreman_oauth_key     = Kafo::Helpers.read_cache_data("oauth_consumer_key")
         foreman_oauth_secret  = Kafo::Helpers.read_cache_data("oauth_consumer_secret")
         katello_oauth_secret  = Kafo::Helpers.read_cache_data("katello_oauth_secret")
-        org                   = @kafo.param('certs', 'org').value
         say <<MSG
 
   To finish the installation, follow these steps:
-
-  If you do not have the smartproxy registered to the Katello instance, then please do the following:
-
-  1. yum -y localinstall http://#{fqdn}/pub/katello-ca-consumer-latest.noarch.rpm
-  2. subscription-manager register --org "<%= color('#{org}', :info) %>"
-
-  Once this is completed run the steps below to start the smartproxy installation:
 
   1. Ensure that the foreman-installer-katello package is installed on the system.
   2. Copy <%= color("#{certs_tar}", :info) %> to the system <%= color("#{capsule_fqdn}", :info) %>
   3. Run the following commands on the capsule (possibly with the customized
      parameters, see <%= color("#{installer_name} --scenario capsule --help", :info) %> and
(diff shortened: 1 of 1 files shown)


stbenjam:
What I don't like, is the org name/location name is hard coded.  Maybe we could do an API query to get a list of all the orgs and suggest the first org and location in the command?

hooks/post/10-post_install.rb
```diff
@@ -15,8 +15,6 @@ def new_install?
   !File.exist?(success_file)
 end
 
-installer_name = kafo.config.app[:installer_name] || kafo.invocation_path
```
ehelms: Looks like you overwrote the previous change to add this

stbenjam: That change wasn't merged yet when I was working on it, and thought we could just get away with removing this.  But I guess you need to overwrite it to say satellite-installer

hooks/post/10-post_install.rb
```diff
@@ -44,23 +44,22 @@ def new_install?
...
   To finish the installation, follow these steps:
 
-  If you do not have the smartproxy registered to the Katello instance, then please do the following:
-
-  1. yum -y localinstall http://#{fqdn}/pub/katello-ca-consumer-latest.noarch.rpm
-  2. subscription-manager register --org "<%= color('#{org}', :info) %>"
-
-  Once this is completed run the steps below to start the smartproxy installation:
-
   1. Ensure that the foreman-installer-katello package is installed on the system.
```
stbenjam: @ehelms I put the `installer_name` back, but this package name is also going to cause a problem.  This would need to say `satellite-installer` downstream

ehelms: Yea, maybe this should be more generic to "make sure the installation instructions for a Capsule have been followed" or something. Given this does not really tell anyone HOW to get that package.

ehelms:
@stbenjam given branch time is approaching, do you want to try to get this in?

stbenjam:
I'll try next week to get the foreman side comments addressed, will let you know

ehelms:
@stbenjam i assume this is still open and waiting on the other PR?

stbenjam:
Yup, there's 2 foreman PR's I need to get merged first

ekohl:
Which PRs is this blocked on?

stbenjam:
@ekohl In description.  I'll come back to it when I have a minute to see if it makes 1.14, but I don't think it's super urgent.

I feel a little uncomfortable about the Foreman changes and the addition of a bunch of API calls to the registration, as this fails pretty often on Katello. Although, I'd guess most of the errors would happen with just 1 or 10 api calls since it's usually network related or the foreman isn't running so maybe it's not a reasonable worry.

ehelms:
This has been open for some time, leaving it as such since the dependent Foreman PRs are still open.

stbenjam:
This isn't going anywhere anytime soon.

PR #345, opened 2016-05-26
Fixes #15205: Add migration creator script

ehelms:
Basic migration generator script that ensure the proper timestamp
formatted file is created in the place you want it.


ekohl:
ACK

PR #346, opened 2016-05-26
Fixes #15203: Set server_ssl_crl to false through migration

ehelms:
In some upgrade scenarios, this field is not being set during the
legacy migration step.


bbuckingham:
ACK!

PR #347, opened 2016-05-26
Fixes #15208: Remove sam-installer

ehelms:


ekohl:
I can't judge if sam-installer is no longer needed, but the change itself looks good :+1:

jlsherrill:
ACK

PR #348, opened 2016-05-27
Fixes #15216 - Add an additional hostname check to that it will fail …

bkearney:
…with a non valid TLD.


stbenjam:
This seems hard to maintain, new TLD's are added all the time.  Why not just regex against having a hyphen in a hostname?

I don't think there's really any restriction on a private network for what you used as a TLD anyway, this will break many many users of Katello who  use a TLD like a 'local', 'intranet', etc.

bkearney:
@stbenjam this was not clear from the bug. Are you saying that:

this.example-name 

is bad, but

this.example_name 

is ok?

If so, I will change the PR to look for the last node not having a -.

stbenjam:
I'm not sure @mccun934 can maybe clarify the point of the bug.

I'm not clear on the bug details, @mccun934 maybe can shed some light? The bug says this:

```
1) Setup a system who's domain name resolves to something with a hyphen:

sat6test.example-domain.com
```

But that's a valid host name I thought.  What were the errors encountered when using a hyphen?

I know we see problems when users try to use underscores in hostnames, e.g. example_name should actually not be valid.

ekohl:
@stbenjam is correct. a hyphen is valid in a hostname, an underscore isn't according to RFCs (but often does work).

bkearney:
We already check for _ in the hostname. The downstream bug is specific about the TLD. I am happy to add "local" to the list as that seems common. @mccun934  thoughts?

stbenjam:
The title says TLD, but the example in the first comment is using a hypen which is legal. 

I wouldn't restrict the TLD, I can imagine that causing a lot of pain for people.  

I thought IANA's canonical list of TLD's is just for the internet? Maybe I'm wrong but I know plenty of users who use all kids of TLD's in their local networks and AFAIK that's allowed.

ekohl:
*If* you want to check if it's a valid TLD, then maybe you should do a DNS-request and check if there are NS records set. Then users can still use a custom TLD, as long as they set it up to resolve properly. The only risk I see is for TLDs that have a SLD (think .co.uk).

stbenjam:
The bugzilla and redmine issue are closed, and this appears stalled.  Going to close this for now, please feel free to reopen if we determine the root cause of the reported issue.  Thanks!

PR #349, opened 2016-05-27
fixes #15218 - adds missing gems for elasticsearch removal

cfouant:


beav:
change is very straightforward, ACK

PR #350, opened 2016-05-27
fixes #14858 - remove gutterball directory

stbenjam:
One small thing I found in testing  upgrade w/ gutterball removal.  The gutterball dir isn't removed.  It's empty, but Tomcat still tries to "deploy" it:

```
May 27 16:39:44 centos7-2-4.example.com server[10839]: INFO: Deploying web application directory /var/lib/tomcat/webapps/gutterball
```


ehelms:
ACK

PR #353, opened 2016-06-01
Fixes #14956 - Check for proxy before katello-installer runs

chris1984:
Created proxy check for katello-installer to verify if a proxy env is set.


stbenjam:
Thanks for your patience @chris1984, tested and works, thanks. Merging.

PR #354, opened 2016-06-02
Fixes #15277 - use `yum erase` instead of `rpm -e`.

beav:
The gutterball removal requests the removal of a handful of packages.
If any are missing with `rpm -e`, the entire command will fail.
However, if you use `yum erase`, the missing packages will not cause an
issue and the remaining packages will be removed. Additionally, `yum
erase` will return a zero exit code in this case.

### DIFF
--- a/hooks/post/30-upgrade.rb
+++ b/hooks/post/30-upgrade.rb
@@ -42,8 +42,12 @@ end
 
 def remove_gutterball
   return true unless Kafo::Helpers.execute('rpm -q gutterball')
-  Kafo::Helpers.execute("rpm -e gutterball tfm-rubygem-foreman_gutterball gutterball-certs tfm-rubygem-hammer_cli_gutterball")
-  Kafo::Helpers.execute("rmdir /var/lib/tomcat/webapps/gutterball")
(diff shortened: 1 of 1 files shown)


beav:
note: i am doing a runthrough of an upgrade with this PR now, it may take some time to complete. The change itself is very small and I hand-tested outside of the full upgrade process.

beav:
patch updated for el6 support on dir removal, testing now on upgrade instance.

beav:
on el6:

```
<snip>
Upgrade Step: remove_gutterball...
Katello upgrade completed!
# rpm -qa | grep gutter

# ls /var/lib/tomcat6/webapps/
candlepin

```
@bbuckingham @ehelms I did not test el7 but it should work the same, the only difference is in the dir name that is now conditionized.

akofink:
@cfouant might also be interested.

bbuckingham:
@beav, The changes look good.  I am currently running a test on my el7 server where I reproduced the issue.  I will let you know when it is finished.

bbuckingham:
@beav, result on el7:
```
<snip>
Upgrade Step: remove_gutterball...
Katello upgrade completed!

# rpm -qa|grep gutter

# ls /var/lib/tomcat/webapps/
candlepin
# 
```

bbuckingham:
ACK

ehelms:
Thanks @beav !

PR #357, opened 2016-06-06
fixes #15315: update capsule foreman url & trusted hosts

bbuckingham:
This commit will ensure that during a capsule upgrade
that the foreman url and trusted hosts are populated
correctly from the legacy capsule configuration.

In addition, a change is being introduced to keep the
default log_dir and log_name from the new configurations
versus from the old (e.g. /v/l/capsule-installer).  This
was the intended behavior and is what is used by some
downstreams.

### DIFF
--- a/hooks/pre_migrations/11-migrate_legacy_katello_config.rb
+++ b/hooks/pre_migrations/11-migrate_legacy_katello_config.rb
@@ -13,7 +13,7 @@ if File.exists?(legacy_config_file)
     scenario = Kafo::Configuration.new(scenario_file)
 
     # copy over config and answers
-    scenario.migrate_configuration(legacy_config)
+    scenario.migrate_configuration(legacy_config, :skip => [:log_dir, :log_name])
     scenario.store(legacy_config.answers)
 
     # link last used scenario
--- a/config/capsule.migrations/160601202256-additional-foreman-proxy.rb
+++ b/config/capsule.migrations/160601202256-additional-foreman-proxy.rb
@@ -1,0 +1,25 @@
+def move(name, default=nil, new_name=nil)
+  answers['foreman_proxy'][(new_name || name)] = answers['capsule'].delete(name) { |k| default }
+end
+
+# migrate from legacy capsule, if exists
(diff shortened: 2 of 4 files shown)


beav:
ACK from me, this worked after also applying a variant of https://github.com/Katello/katello-packaging/pull/233

hooks/pre_migrations/11-migrate_legacy_katello_config.rb
```diff
@@ -13,7 +13,7 @@
     scenario = Kafo::Configuration.new(scenario_file)
 
     # copy over config and answers
-    scenario.migrate_configuration(legacy_config)
+    scenario.migrate_configuration(legacy_config, :skip => [:log_dir, :log_name])
```
mbacovsky: I'm not sure if users tend to touch the logging setup but it is the case we could do conditional migration if the value differs from our old defaults:
```ruby
scenario.app[:log_dir] = legacy_config.app[:log_dir] unless legacy_config.app[:log_dir] == '/var/log/katello-installer'
scenario.app[:log_name] = legacy_config.app[:log_name] unless legacy_config.app[:log_name] == 'katello-installer.log'
```

mbacovsky:
I'm not sure if it makes sense to do the change I describeed inline. LGTM either way.

bbuckingham:
@beav, @mbacovsky, I updated the migration to include moving some additional parameters from the legacy configuration.

@mbacovsky, per our offline chat, the plan for logging at the moment is to transition to the new logging location (/var/log/foreman-installer).  This will ensure that various tools (such as foreman-debug) are able to collect the logs, when needed.

beav:
@bbuckingham the change looks good but I will retest on el6 and report back

mbacovsky:
@bbuckingham, that is reasonable. The new update looks good. (the move function is a good candidate for moving into migration context in kafo in next version).
ACK from me. :beer:

beav:
ACK, tested on el6

bbuckingham:
@beav & @mbacovsky,  Thanks!

PR #358, opened 2016-06-07
Fixes #15325 - Handle nil params

akofink:
for 'capsule_certs' and 'parent_fqdn'

BATs are failing with `/usr/share/katello-installer-base/hooks/post/10-post_install.rb:22:in `block (4 levels) in load': undefined method `value' for nil:NilClass (NoMethodError)`

This was introduced in #332.

### DIFF
--- a/hooks/post/10-post_install.rb
+++ b/hooks/post/10-post_install.rb
@@ -19,7 +19,7 @@ installer_name = kafo.config.app[:installer_name] || kafo.invocation_path
 
 if [0,2].include?(@kafo.exit_code)
   if !app_value(:upgrade)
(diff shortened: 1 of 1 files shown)


hooks/post/10-post_install.rb
```diff
@@ -19,7 +19,7 @@ def new_install?
 
 if [0,2].include?(@kafo.exit_code)
   if !app_value(:upgrade)
-    fqdn = @kafo.param('capsule_certs','parent_fqdn').value || Facter.value(:fqdn)
+    fqdn = @kafo.param('capsule_certs','parent_fqdn').try(&:value) || Facter.value(:fqdn)
```
jlsherrill: I think you might mean   `.try(:value)`

jlsherrill:
ACK merging!

PR #359, opened 2016-06-07
Fixes #15327 - Handle nil params

akofink:
Revert "Fixes #15325 - Handle nil params"

This reverts commit 0aa563937629ca8ab66c76e4705b8c917b927550.

### DIFF
--- a/hooks/post/10-post_install.rb
+++ b/hooks/post/10-post_install.rb
@@ -19,7 +19,11 @@ installer_name = kafo.config.app[:installer_name] || kafo.invocation_path
 
 if [0,2].include?(@kafo.exit_code)
   if !app_value(:upgrade)
-    fqdn = @kafo.param('capsule_certs','parent_fqdn').try(:value) || Facter.value(:fqdn)
+    fqdn = if @kafo.param('capsule_certs','parent_fqdn')
+             @kafo.param('capsule_certs', 'parent_fqdn').value
+           else
+             Facter.value(:fqdn)
+           end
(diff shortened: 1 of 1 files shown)


stbenjam:
I'd just use this PR to fix the actual problem instead of reverting + fixing later, to reduce churn in the code here.

`@kafo.param('capsule_certs','parent_fqdn') ?  @kafo.param('capsule_certs','parent_fqdn').value : Facter.value(:fqdn)` should do it

akofink:
I think we want to try and figure out why `@kafo.param('capsule_certs', 'parent_fqdn')` is nil, since we use this same paradigm all over the place without a nil check.

stbenjam:
I think it's because the `capsule_certs` answer only exists on capsule-certs-generate, so it's nil on the katello and capsule scenarios.  Whereas everywhere else we call param values, the class and param always exists in every case where it's called.

akofink:
Here's another place where we're using these parameters. We check for nil first: https://github.com/Katello/katello-installer/blob/83f11669c8d9f9dac2fe7434831f94f2851e8146/hooks/pre/20-certs_update.rb#L31

Alright, let's just check for nil and move on.

akofink:
@stbenjam I implemented your fix but in an if/else since multi-line ternary operators are considered bad style (by rubocop in our other projects).

stbenjam:
I agree but 116 characters is far from the longest line in this file :laughing: 

Anywho, ACK

PR #360, opened 2016-06-08
Fixes #15346 - Guess what is running from scenario name

mbacovsky:
This patch should allow us to be more precise about what is running in post install message. E.g.:
```text
[root@sat-snap-rhel7 ~]# foreman-installer --scenario satellite
Installing             Done                                               [100%] [...............................................................................................................................]
  Success!
  * Satellite is running at https://sat-snap-rhel7.example.com

  The full log is at /var/log/foreman-installer/satellite.log
```

### DIFF
--- a/bin/capsule-certs-generate
+++ b/bin/capsule-certs-generate
@@ -31,6 +31,7 @@ CONFIG_CONTENT = <<EOF
 :answer_file: #{answers_file.path}
 :installer_dir: #{INSTALLER_DIR}
 :installer_name: 'foreman-installer'
+:product_name: 'Katello'
 :module_dirs: ["/usr/share/foreman-installer/modules", "#{INSTALLER_DIR}/modules"]
 :hook_dirs: ["#{INSTALLER_DIR}/hooks"]
 :default_values_dir: /tmp
--- a/config/capsule.yaml
+++ b/config/capsule.yaml
@@ -1,6 +1,7 @@
 ---
   :name: Capsule
   :description: Install a stand-alone Capsule.
+  :product_name: Katello
   :log_dir: /var/log/foreman-installer
   :log_name: capsule.log
   :log_level: DEBUG
--- a/config/katello-devel.yaml
+++ b/config/katello-devel.yaml
@@ -1,6 +1,7 @@
 ---
   :name: Katello devel
   :description:
+  :product_name: Katello
   :log_dir: /var/log/foreman-installer
   :log_name: katello-devel.log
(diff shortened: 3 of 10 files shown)


ehelms:
Might as well do it here too:

https://github.com/Katello/katello-installer/pull/360/files#diff-d9b1058a236a3d65bd0c4fbc0b60f640R57

mbacovsky:
@ehelms, thanks, I missed that one. But when I tried to find in which scenario the 'capsule_certs' module is used I didn't found any. Is this module still used or was it renamed? I also didn't noticed the message during the Sat installation.

mbacovsky:
@ehelms, I didnt found a way to refer to 'Katello' using existing data so I decided to extend installer configuration with `product_name` and use that. I replaced all Katello references and tested the patch with downstream and branding there. I'd be happy for any feedback.

bin/capsule-certs-generate
```diff
@@ -31,6 +31,7 @@ CONFIG_CONTENT = <<EOF
 :answer_file: #{answers_file.path}
 :installer_dir: #{INSTALLER_DIR}
 :installer_name: 'foreman-installer'
+:product_name: 'Katello'
```
ehelms: `product_name` seems a bit on the nose, since we are dealing with scenarios -- why not use the name defined in the config file for the scenario?

mbacovsky: Scenario name is not IMO suitable for this, because if we need to refer to Katello server from Capsule scenario the name will be Capsule. I'm not very happy with 'product_name' too. Maybe `programm_name` or `server_name` would be better. Also moving the value to the `custom:` section could help.

ehelms: I see your point; however, I don't see an instance where this is the case looking through the PR. If you did need to refer to Katello during a Capsule install, you'd likely need to refer to the Capsule as well which would mean you'd need to define product_name twice?

mbacovsky: There can be problem in `capsule-certs-generate` (https://github.com/Katello/katello-installer/blob/master/hooks/post/10-post_install.rb#L56) which is implemented using `kafo`. There the config has no scenario name and adding there scenario name 'Satellite' seems to be hackish. 

I agree about the `product_name`. For the capsule example it would make more sense to have `server_name` or `server_product_name` and `proxy_name` (`proxy_product_name`) or something similar.

mbacovsky:
I'm closing the PR. I'll reopen updated one when the underlying work in Kafo is ready.

PR #361, opened 2016-06-08
fixes #15350: add migrations for capsule -> foreman_proxy separation

bbuckingham:
With the Katello on Foreman feature, there was an effort to separate
the foreman_proxy aspects from the puppet-capsule.  The changes
for this were primarily done as part of:
	https://github.com/Katello/puppet-capsule/pull/64

At that time, there were no migration capabilities to move the
answers from the 'capsule' to 'foreman_proxy' within the various
answer files (e.g. answers.katello-installer.yaml &
answers.capsule-installer.yaml).

This commit will add in migrations for each env to support moving
the parameters over.  This includes:
- katello
- katello-devel
- capsule

It should be noted that for katello and katello-devel include
an additional migration.  It is one that was added in a previous
commit for capsule, but it should apply to all environments.

### DIFF
--- a/config/capsule.migrations/160608111430-additional-foreman-proxy.rb
+++ b/config/capsule.migrations/160608111430-additional-foreman-proxy.rb
@@ -1,0 +1,44 @@
+def move(name, new_name=nil, default=nil)
+  answers['foreman_proxy'][(new_name || name)] = answers['capsule'].delete(name) { |k| default }
+end
+
+# migrate addtional fields from the legacy capsule configuration to foreman_proxy
+if answers['capsule'].is_a? Hash
+  move('bmc', 'bmc')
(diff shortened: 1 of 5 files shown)


beav:
this worked for me sat-side, testing capsule now

beav:
This worked for me, ACK

bbuckingham:
@beav, Thanks for the review and testing on el6.  I tested the patch on el7 as well and the behavior was as I expected. 

I am going to leave the PR open overnight, in case there are other reviewers.  If all goes well, we can merge some time tomorrow.

mbacovsky:
ACK. Thanks @bbuckingham!

bbuckingham:
@beav, @mbacovsky, Thanks guys!  Merging... :)

PR #362, opened 2016-06-08
Fixes #15351: Add missing upgrade rake tasks

ehelms:


ehelms:
@jlsherrill please review

hooks/post/30-upgrade.rb
```diff
@@ -36,6 +36,10 @@ def elasticsearch_message
   Kafo::Helpers.log_and_say :info, message
 end
 
+def add_export_distributor
+  Kafo::Helpers.execute('foreman-rake katello:upgrades:3.0:add_export_distributor')
+end
```
ehelms: @beav note that I noticed this never got added as an upgrade rake task and included it here

jlsherrill:
ack

hooks/post/30-upgrade.rb
beav: I originally added this when I was pretty new to the team, thanks for catching this :panda_face:

PR #364, opened 2016-06-13
Fixes #15390: Ignore empty answers during Capsule to Proxy migration

ehelms:
Without this check, answers not found already (such as a new
installation) are set to nil instead of the default values provided
by the puppet module.

### DIFF
--- a/config/katello.migrations/160608111430-additional-foreman-proxy.rb
+++ b/config/katello.migrations/160608111430-additional-foreman-proxy.rb
@@ -1,4 +1,5 @@
 def move(name, new_name=nil, default=nil)
+  return unless answers['capsule'].key?(name)
   answers['foreman_proxy'][(new_name || name)] = answers['capsule'].delete(name) { |k| default }
 end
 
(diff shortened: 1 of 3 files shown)


config/katello.migrations/160608111430-additional-foreman-proxy.rb
```diff
@@ -1,4 +1,5 @@
 def move(name, new_name=nil, default=nil)
+  return if answers['capsule'].key?(name)
```
bbuckingham: Would we need similar statement in 160120212701-foreman-proxy.rb & 160601202256-additional-foreman-proxy.rb ?

ehelms: I don't think so... however, it probably would not hurt for me to add it there just to be safe.

ehelms:
@bbuckingham updated to add it to all three

bbuckingham:
ACK.  Thanks @ehelms !

PR #365, opened 2016-06-13
fixes #15393 - add foreman hiera config to katello scenarios

stbenjam:


ekohl:
This relies on the foreman-installer package being installed, right? I don't have a great view on whether this works in all cases.

stbenjam:
I'm not quite sure if this works in all cases, maybe it should be a relative path and do a sed replace in the rpm spec like we do for other settings: https://github.com/Katello/katello-packaging/blob/master/katello-installer/katello-installer-base.spec#L97

stbenjam:
But even on a development install, we now actually install the foreman-installer package itself

ehelms:
The pathing should be OK in all cases with what we support and how we handle things right now. I think we will need an equivalent https://github.com/theforeman/foreman-installer/commit/e5dce673886caee6bebfc0ad91e19d43bbabb737#diff-52b611d4a709b477a55270b5ca1c11c3

stbenjam:
Do you mean to include our own separate hiera config? It's going to be a pain if we're maintaining one for each of our scenarios and always have to import their changes isn't it?

With this PR I get a clean install, since I'm referencing the foreman hiera config directly.We probably want to always take Foreman's hiera config, don't we?

ehelms:
Bleh, that link didn't translate well to what I wanted to point you to. We'll need the migration to add this to existing scenario configuration files right? I am all for not maintaining the hiera config part :)

stbenjam:
Oh right, of course, updated

ehelms:
Install passed -- ACK!

stbenjam:
Thanks!

PR #366, opened 2016-06-15
Refs #15390: Ignore empty answers when moving

ehelms:


ehelms:
@bbuckingham I should have thought of this when you pointed out the other cases in the katello.migrations

bbuckingham:
ACK

PR #370, opened 2016-06-23
Fixes #15511 - fix setup_local to copy to correct dir

jlsherrill:


jlsherrill:
without this, when running the rake task it put the moduels in ./modules/modules/ which didn't seem correct, unless I'm just doing it wrong

stbenjam:
Ah that makes sense if the modules directory already exists, but what if the modules directory isn't empty like when someone ran librarian manually? Maybe it might make sense to delete the local modules directory, too, before the cp_r.

jlsherrill:
ahhh!  that makes sense.  I will update to do that

stbenjam:
Looks ok to me, ACK

PR #373, opened 2016-07-12
Fixes #15660: Print command output during upgrade steps in debug color

ehelms:
This is especially useful for Pulp migrations as they contain useful
information regarding what stage they are at and if they will be long
running.

### DIFF
--- a/hooks/boot/01-helpers.rb
+++ b/hooks/boot/01-helpers.rb
@@ -7,7 +7,15 @@ class Kafo::Helpers
     end
 
     def log_and_say(level, message)
-      style = level == :error ? 'bad' : level
+      style = case level
+              when :error
+                'bad'
+              when :debug
+                'yellow'
+              else
+                level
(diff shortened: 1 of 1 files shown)


hooks/boot/01-helpers.rb
```diff
@@ -8,6 +8,7 @@ def module_enabled?(kafo, name)
 
     def log_and_say(level, message)
       style = level == :error ? 'bad' : level
+      style = :yellow if level == :debug
```
stbenjam: Do you think a case statement is clearer? Something like

```ruby
style = case level
        when :error
          'bad'
        when :debug
          'yellow'
        else
          level
        end
```

stbenjam:
But otherwise :+1: :+1:

ehelms:
@stbenjam updated with the case statement

stbenjam:
ACK

PR #374, opened 2016-07-14
Fixes #15683 - escape output before printing to user

iNecas:


ehelms:
ACK - thanks @iNecas !

PR #375, opened 2016-07-15
Fixes #15696 - Added memory check to installer

chris1984:
Added memory check to installer to make sure system has decent ram before running the installer.

### DIFF
--- a/hooks/pre/17-memory_check.rb
+++ b/hooks/pre/17-memory_check.rb
@@ -1,0 +1,11 @@
+# call grep and awk to store total ram and define min ram
+total_ram = `grep MemTotal /proc/meminfo | awk '{print $2}'`.to_i
+min_ram = 8_388_608
+
+# call mem_check if flag is called
+if app_value(:disable_system_checks)
+  logger.warn 'Skipping system checks.'
+elsif min_ram > total_ram
+  $stderr.puts 'This system has less than 8 GB of total memory. Please have at least 8 GB of total ram free before running the installer.'
+  kafo.class.exit(1)
+end
--- a/hooks/boot/17-disable_system_checks.rb
+++ b/hooks/boot/17-disable_system_checks.rb
@@ -1,0 +1,7 @@
+# Add disable system checks option
+app_option(
+  '--disable-system-checks',
+  :flag,
+  "This option will skip the system checks for memory.",
+  :default => false
+)


iNecas:
There should be still a way at least to force skip this check: As a developer, I need to run more than one VM sometimes, and I can take the responsibility for the memory allocation and that it might not be enough for some cases.

stbenjam:
Unfortunately, `boot` (where we add options) runs after anything in checks.  We'll have to be a little creative.  Either read an environment variable, or use the option anyway with a little creativity (check something like `ARGV.incldue?('--disable-memory-check')` instead of using the app_value helper).

sideangleside:
Also, would it be a better idea to use facter to get the amount of installed memory? That way you don't have to parse the output of the `free` command. 

~~~
begin
  require 'rubygems'
rescue LoadError
end
require 'facter'

total_ram = Facter.value(:memorysize_mb)
~~~

chris1984:
@stbenjam updated with a env variable check here is my output from my testing

chris1984:
without variable

[root@centos7-katello-3-1 ~]# foreman-installer --scenario katello -v
This system has less than 8 GB of total memory. Please have at least 8 GB of total ram free before running the installer.
Your system does not meet configuration criteria
[ INFO 2016-08-25 11:47:54 verbose] Executing hooks in group pre_migrations
[ INFO 2016-08-25 11:47:54 verbose] All hooks in group pre_migrations finished


With variable:

[root@centos7-katello-3-1 ~]# export DISABLE_INSTALLER_CHECKS="true"
[root@centos7-katello-3-1 ~]# echo $DISABLE_INSTALLER_CHECKS
true

[root@centos7-katello-3-1 ~]# foreman-installer --scenario katello -v
DISABLEDCHECK
[ INFO 2016-08-25 11:48:49 verbose] Executing hooks in group pre_migrations
[ INFO 2016-08-25 11:48:49 verbose] All hooks in group pre_migrations finished
[ INFO 2016-08-25 11:48:49 verbose] Executing hooks in group boot
[ INFO 2016-08-25 11:48:49 verbose] All hooks in group boot finished
[ INFO 2016-08-25 11:48:49 verbose] Executing hooks in group init
[ INFO 2016-08-25 11:48:49 verbose] All hooks in group init finished
[ INFO 2016-08-25 11:50:25 verbose] Executing hooks in group post
  Success!
  * Katello is running at https://centos7-katello-3-1.example.com

[ INFO 2016-08-25 11:50:25 verbose] All hooks in group post finished

chris1984:
@sideangleside updated to use facter :)

stbenjam:
We don't use environment variables anywhere else that I know of, I think it's better if you use installer params to bypass the check, plenty of examples in the boot directory, for example;

https://github.com/Katello/katello-installer/blob/master/hooks/boot/30-upgrade.rb

And you check the value with `app_value(:upgrade)`

checks/memory.rb
```diff
@@ -0,0 +1,40 @@
...
+# Check for the presence of the DISABLE_INSTALLER_CHECKS
+# Environment Variable.  If it is nil, we dont want
+# to disable checking. Otherwise TRUE = nochecks
+# FALSE (or any other value) = check
+
+begin
+  require 'rubygems'
+rescue LoadError
+end
+require 'facter'
```
stbenjam: You cannot use the facter gem directly, this is what we were talking about on IRC.  In Puppet 4, the gem isn't accessible to the installer.  You would need to shell out and call facter binary or use the original version you had.

hooks/pre/17-memory_check.rb
```diff
@@ -0,0 +1,20 @@
...
+min_ram = 8388608
+       
+# call mem_check if flag is called
+if app_value(:disable_system_checks) then
+logger.warn 'Skipping system checks.'
+else
+ if min_ram > total_ram
+logger.error 'This system has less than 8 GB of total memory. Please have at least 8 GB of total ram free before running the installer.'
+kafo.class.exit 1
+ end
```
stbenjam: Please fix indentation, all ruby code needs to follow [our style guide](http://www.katello.org/developers/style/ruby.html)

hooks/boot/21-disable_system_checks.rb
```diff
@@ -0,0 +1,7 @@
+# Add reset option
+app_option(
+  '--disable_system_checks',
```
stbenjam: Please use dashes not underscores so it's consistent with all other installer options

stbenjam:
Looks great!! One tiny nitpick, can you make the number prefixes of the file in `boot/` and the file in `pre/` the same? The rest all match up.

stbenjam:
Thanks, looks great, ACK!

chris1984:
@stbenjam can I get an updated check when you get a free moment

PR #377, opened 2016-07-19
Fixes #13546 - Enable TFTP during install

chris1984:


config/katello.migrations/160719114223-enableftp.rb
```diff
@@ -0,0 +1,7 @@
+# enable tftp service for foreman-proxy fixes #13546
+
+# foreman_proxy defaults
+answers['foreman_proxy'] = {
+  'tftp' => true,
+}
```
stbenjam: This probably doesn't do what you intended - it will erase all of the users' other answers for foreman_proxy.

Best to do something like:

```ruby
answers['foreman_proxy']['tftp'] = true if answers['foreman_proxy'].is_a?(Hash)
```

stbenjam:
These migrations are really tricky, this migration enables TFTP on *ALL* Katello's, even if a user explicitly disabled it before.

I think maybe we should just delete this line:
https://github.com/Katello/katello-installer/blob/master/config/katello.migrations/160120212701-foreman-proxy.rb#L38

And change this to true:
https://github.com/Katello/katello-installer/blob/master/config/katello.migrations/160120212701-foreman-proxy.rb#L60

But we have to be careful when changing existing migrations.  But I think this will do what we want:

1. On an existing katello that doesn't been upgraded yet, they'll get their original option.  
2. On a new katello, they'll get tftp set to the default value (true).
3. On a katello that was already upgraded, well, we already turned it off on them even if they wanted it on so we'll just have to deal with that decision

ehelms:
@chris1984 we are nearing branch time for release -- do you plan to revisit this before then?

chris1984:
@ehelms will submit an updated commit based on stbenjam's comments today.

chris1984:
Finishing up testing

chris1984:
test

PR #379, opened 2016-07-21
Fixes #15775 - Use absolute paths instead of relative.

JacobCallahan:
Switching to absolute paths to avoid issues with validate_absolute_path
during upgrades.
Fixes #15775

New format
--------------
```
...
Validation succeeded.

To install the Satellite main server with the custom certificates, run:

    satellite-installer --scenario satellite\
                        --certs-server-cert "/root/sat_cert/test.crt"\
                        --certs-server-cert-req "/root/sat_cert/test.crt.req"\
                        --certs-server-key "/root/sat_cert/test.key"\
                        --certs-server-ca-cert "/root/sat_cert/cacert.crt"
...
```

### DIFF
--- a/bin/katello-certs-check
+++ b/bin/katello-certs-check
@@ -12,16 +12,16 @@ HELP
 while getopts "c:k:r:b:" opt; do
     case $opt in
         c)
-            CERT_FILE="$OPTARG"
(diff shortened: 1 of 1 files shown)


ehelms:
ACK! Thanks @JacobCallahan !

iNecas:
Nice!

PR #381, opened 2016-07-29
fixes #15911 - don't use facter anymore in the installer

stbenjam:
Facter is bundled in the AIO bundle and not accessible to system ruby.  We can just call `hostname -f` anyway.


beav:
ACK

PR #382, opened 2016-08-01
fixes #15938 - generate parser cache for kafo

komidore64:
depends on theforeman/kafo#166.

### DIFF
--- a/config/katello.yaml
+++ b/config/katello.yaml
@@ -9,6 +9,9 @@
   :installer_dir: "."
   :default_values_dir: "/tmp"
   :module_dirs: ["./_build/modules", "../katello-installer/modules"]
+  :parser_cache_path:
+    - /usr/share/foreman-installer/parser_cache/foreman.yaml
+    - /usr/share/katello-installer-base/parser_cache/katello.yaml
   :hiera_config: /usr/share/foreman-installer/config/foreman-hiera.conf
   :hook_dirs: ["../katello-installer/hooks"]
   :default_values_dir: "./config"
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-gem 'kafo', '>= 0.7.1' 
+gem 'kafo', '>= 0.9.3'
 gem 'rake'
 gem 'puppet', '>= 3.6.0', '< 4.0.0'
 gem 'librarian-puppet', '~> 2.2.0'
--- a/Rakefile
+++ b/Rakefile
@@ -1,5 +1,7 @@
 require 'rake/clean'
 require 'tempfile'
+require 'kafo'
+require 'yaml'
 
 BUILDDIR = File.expand_path(ENV['BUILDDIR'] || '_build')
 PKGDIR = ENV['PKGDIR'] || File.expand_path('pkg')
@@ -17,10 +19,21 @@ end
 file "#{BUILDDIR}/modules" do |t|
   if Dir["modules/*"].empty?
     sh "librarian-puppet install --verbose --path #{BUILDDIR}/modules"
-    Rake::Task["prune_foreman_modules"].execute
   else
     mkdir BUILDDIR
     cp_r "modules", "#{BUILDDIR}/modules"
+  end
+end
+
+file "#{BUILDDIR}/parser_cache/katello.yaml" => [BUILDDIR] do |filename|
+  mkdir "#{BUILDDIR}/parser_cache"
+  sh "kafo-export-params -c config/katello.yaml -f parsercache --no-parser-cache -o #{filename}"
+
+  # strip out forman bits from the parser cache
+  cache = YAML.load_file(filename.to_s)
+  cache[:files] = cache[:files].delete_if { |k, _| k =~ /\Aforeman/ }
+  File.open(filename.to_s, "w") do |file|
+    file.write(cache.to_yaml)
   end
 end
 
@@ -49,19 +62,20 @@ end
 
 namespace :pkg do
   desc 'Generate package source tar.bz2'
-  task :generate_source => [:clean, PKGDIR, "#{BUILDDIR}/modules"] do
+  task :generate_source => [:clean, PKGDIR, "#{BUILDDIR}/modules", "#{BUILDDIR}/parser_cache/katello.yaml", :prune_foreman_modules] do
 
     version = File.read('VERSION').chomp
     raise "can't read VERSION" if version.length == 0
 
     Dir.chdir(BUILDDIR) do
-      puts "tar -cf #{BUILDDIR}/modules.tar --transform=s,^,katello-installer-#{version}/, modules/"
-      `tar -cf #{BUILDDIR}/modules.tar --transform=s,^,katello-installer-#{version}/, modules/`
+      sh "tar --create --file=#{BUILDDIR}/modules.tar --transform=s,^,katello-installer-#{version}/, modules/"
+      sh "tar --create --file=#{BUILDDIR}/parser_cache.tar --transform=s,^,katello-installer-#{version}/, parser_cache/"
     end
 
-    `git archive --prefix=katello-installer-#{version}/ HEAD > #{PKGDIR}/katello-installer-#{version}.tar`
-    `tar --concatenate --file=#{PKGDIR}/katello-installer-#{version}.tar #{BUILDDIR}/modules.tar`
-    `gzip -9 #{PKGDIR}/katello-installer-#{version}.tar`
+    sh "git archive --prefix=katello-installer-#{version}/ HEAD > #{PKGDIR}/katello-installer-#{version}.tar"
+    sh "tar --concatenate --file=#{PKGDIR}/katello-installer-#{version}.tar #{BUILDDIR}/modules.tar"
+    sh "tar --concatenate --file=#{PKGDIR}/katello-installer-#{version}.tar #{BUILDDIR}/parser_cache.tar"
+    sh "gzip -9 #{PKGDIR}/katello-installer-#{version}.tar"
   end
 end
 
--- a/config/capsule.migrations/160824145544-parser_cache.rb
+++ b/config/capsule.migrations/160824145544-parser_cache.rb
@@ -1,0 +1,4 @@
+scenario[:parser_cache_path] ||= [
+  '/usr/share/foreman-installer/parser_cache/foreman.yaml',
+  '/usr/share/katello-installer-base/parser_cache/katello.yaml'
+]
--- a/config/capsule.yaml
+++ b/config/capsule.yaml
@@ -9,6 +9,9 @@
   :installer_dir: "."
   :default_values_dir: "/tmp"
   :module_dirs: ["/usr/share/foreman/modules", "./modules"]
+  :parser_cache_path:
+    - /usr/share/foreman-installer/parser_cache/foreman.yaml
+    - /usr/share/katello-installer-base/parser_cache/katello.yaml
   :hiera_config: /usr/share/foreman-installer/config/foreman-hiera.conf
   :hook_dirs: ["./hooks"]
   :colors: true
--- a/config/katello-devel.migrations/160824145544-parser_cache.rb
+++ b/config/katello-devel.migrations/160824145544-parser_cache.rb
@@ -1,0 +1,4 @@
+scenario[:parser_cache_path] ||= [
+  '/usr/share/foreman-installer/parser_cache/foreman.yaml',
+  '/usr/share/katello-installer-base/parser_cache/katello.yaml'
+]
--- a/config/katello-devel.yaml
+++ b/config/katello-devel.yaml
@@ -9,6 +9,9 @@
   :installer_dir: "."
   :default_values_dir: "/tmp"
   :module_dirs: ["./_build/modules", "../katello-installer/modules"]
+  :parser_cache_path:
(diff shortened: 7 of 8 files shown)


stbenjam:
Please update Gem requirement to Kafo >= 0.8

komidore64:
I guess you could say this is testable by running `rake` to generate a source tarball in `pkg/`, then ``tar tf pkg/katello-installer*.tar.gz`` to confirm the parser cache is included.

stbenjam:
The error on travis I get locally, too:

```
Manifest foreman/manifests/init.pp was not found in /home/stbenjam/git/katello-installer/_build/modules, /home/stbenjam/git/katello-installer/modules
/home/stbenjam/.rvm/gems/ruby-2.0.0-p643/gems/kafo-0.9.2/lib/kafo/puppet_module.rb:43:in `join': no implicit conversion of nil into String (TypeError)
```

It's because we prep the modules by running the `setup_local` rake task installs the modules and then removes the foreman ones (we don't want to package them).   Maybe @ares has an idea how to avoid this in Kafo, we really only want to store the parser cache for our modules, including foreman could cause problems as there might be different versions of the modules in the RPM package than what we generate the cache for.

komidore64:
> Please update gem requirement to Kafo >= 0.8

thanks, updated!

ares:
I'm probably missing something. You can generate the cache the cache for whole module set only, so you need Foreman modules I think. Maybe some sort of cache merging could be added to Kafo. Otherwise you need the same versions of Foreman modules that foreman-installer uses. If you don't want to package them and let foreman-installer install them, you should probably clean those as part of packaging, not listing them in %files?

ehelms:
@ares The katello-installer only includes the modules that it uses and are a subset of all the modules required. That is to say we prune out any module provided by foreman-installer from our modules when we generate the tarball. Thus, the configuration points over to where the foreman-installer installs modules and where our installer installs modules [1]. This in theory means the versions of some modules could be slightly different depending on build time of each package. If we generate a parser cache based on all the modules present before pruning we could have erroneous data.


[1] https://github.com/Katello/katello-installer/blob/master/config/katello.yaml#L11

ares:
So is there a chance to have foreman-installer as a build dependency so you'd use its modules during packaging? You could still package only the subset of modules that katello uses I believe.

ehelms:
@ares yes - are you suggesting we include foreman-installer as a build dependency and then generate the parser cache in the RPM pointing at the combined module set? Does the parser cache generation look at the multiple locations for modules pointed to by the Kafo config file (e.g. `modules: [/path/a, /path/b]`) ?

ehelms:
Also, what happens when foreman-installer updates (and modules with it) ? For example, say Foreman 1.11 is released and Katello builds against 1.11.0. Then Foreman releases 1.11.1 which includes module updates but we do not rebuild our installer. Will this cause parser cache differences or weirdness?

ekohl:
To me it sounds like Kafo needs to be taught about separate cache files for the reasons @ehelms mentions. You ship multiple sets of modules so you also need multiple caches for those sets. The responsibility of creating the cache lies with the package that ships them.

ares:
> are you suggesting we include foreman-installer as a build dependency and then generate the parser cache in the RPM pointing at the combined module set?

yes 

> Does the parser cache generation look at the multiple locations for modules pointed to by the Kafo config file (e.g. modules: [/path/a, /path/b]) ?

I think so, the exporter users the kafo configuration object, calling `@config.modules` which is driven by values in answer file specified in config file.

> Also, what happens when foreman-installer updates (and modules with it) ?

good question, I think foreman-installer bump should imply katello-installer bump then. At the end of the day, change in foreman-installer means also change in katello scenario. Maybe we could merge both installers one day.

The same question would remain if the cache could be merged I think. The only option left I see is dropping the cache and enforce puppet-strings module to be installed for puppet 4 installations. We wanted to avoid it so the cache was introduced.

ekohl:
> good question, I think foreman-installer bump should imply katello-installer bump then. At the end of the day, change in foreman-installer means also change in katello scenario. Maybe we could merge both installers one day.

I consider this a long term goal, but I don't see this happening within a year.

ehelms:
@ares how could we do cache per module? is that a kafo update? Even if we merge both installers one day, there would still be the scenario issue I feel like whereby a scenario could have it's own set of modules that it is bringing to the party.

ares:
> how could we do cache per module? is that a kafo update?

@ehelms yes, that means a kafo update. One part is splitting the cache when it's generated, second is merging when it's used.

>  I feel like whereby a scenario could have it's own set of modules that it is bringing to the party.

that's would still be possible, each scenario has it's own cache, it's more about using the same module versions when they are shared among scenarios

ehelms:
@ares do you mind pointing us to the code where this all takes place?

ares:
You can find the original PR at https://github.com/theforeman/kafo/pull/135
most important files are https://github.com/theforeman/kafo/blob/master/lib/kafo/parser_cache_writer.rb and https://github.com/theforeman/kafo/blob/master/lib/kafo/parser_cache_reader.rb

komidore64:
To summarize our options, as far as I understand it.

1. Install ``puppet-strings`` and ``yard`` with the installer and don't worry about a parser cache at all.

2. Include ``foreman`` and ``katello`` modules in ``katello-installer`` to generate our own parser-cache, separate from the cache generated in ``foreman-installer``.

3. Modify ``kafo`` to accept multiple parser caches using some kind of merge strategy.

komidore64:
update:

went with [#3](https://github.com/theforeman/kafo/pull/166).

komidore64:
theforeman/kafo#166 has been merged and a new kafo gem ([`0.9.3`](https://rubygems.org/gems/kafo/versions/0.9.3)) has been cut.

config/katello.yaml
```diff
@@ -9,6 +9,9 @@
   :installer_dir: "."
   :default_values_dir: "/tmp"
   :module_dirs: ["./_build/modules", "../katello-installer/modules"]
+  :parser_cache_path:
+    - /usr/share/foreman-installer/parser_cache/foreman.yaml
+    - ./_build/parser_cache/katello.yaml
```
ehelms: I think if we set this to the following we avoid having to change this during RPM build time:

```
../katello-installer/parser_cache/katello.yaml
```

ehelms: I think we would also need to add this to the devel and capsule scenarios config.

komidore64: > I think if we set this to the following we avoid having to change this during RPM build time:
> 
> ../katello-installer/parser_cache/katello.yaml

would you prefer i didn't stay consistent with the pathing structure for katello-installer?

komidore64:
all green!

config/katello.yaml
ehelms: I think that is more consistent/better.

ehelms: If you note, this is the only other place that uses `_build` [1] and I believe it was there to aide with local testing. But you will notice that the second argument in the array is more in line with my suggestion. And keeps things universally working.

[1] https://github.com/Katello/katello-installer/pull/382/files#diff-f7223b52c32ec607907f68076d4a0db3R11

ehelms:
We will also need migrations similar to [1].


[1] https://github.com/theforeman/foreman-installer/pull/172/files#diff-efa771da30c5f26098a3aebb72dc9f64

komidore64:
updated with parser cache paths for capsule and katello-devel scenarios as well as migrations for the katello, katello-devel, and capsule scenarios.

ehelms:
Scratch build (untested yet) with these changes -- http://koji.katello.org/koji/taskinfo?taskID=517932

komidore64:
I can confirm that the parser cache is located within that katello-installer-base scratch-built RPM, contains the proper contents, and gets installed to `/usr/share/katello-installer-base/parser_cache/katello.yaml`.

beav:
Do I need to tell kafo where the cache is somehow? When I installed the scratch RPM and uninstalled puppetlabs-strings, I got the usual "could not find parser cache" error.

komidore64:
@beav the scenario file tells kafo where to find the parser cache. so i'm pretty sure you need `foreman-installer-katello` along with `katello-installer-base`.

ehelms:
The scratch build appears to not be "newer" than whats in nightly so I am working on a new one to avoid work arounds trying to use it for testing.

ehelms:
Better scratch build (especially if using --koji-task option in Forklift):

http://koji.katello.org/koji/taskinfo?taskID=518316

beav:
@ehelms testing now with `--koji-task`, will let you know result

beav:
@ehelms I am still having some trouble with this.

```bash
# rpm -Uvh 'http://koji.katello.org/koji/getfile?taskID=518316&name=katello-installer-base-3.2.0-2.nightly.git.0.4cbd175.el7.noarch.rpm' 'http://koji.katello.org/koji/getfile?taskID=518316&name=foreman-installer-katello-3.2.0-2.nightly.git.0.4cbd175.el7.noarch.rpm'

# foreman-installer --help --scenario katello
/usr/share/gems/gems/kafo-0.9.4/lib/kafo/puppet_module.rb:69:in `parse': No Puppet module parser is installed and no cache of the file /usr/share/katello-installer-base/modules/certs/manifests/init.pp is available. Please check debug logs and install optional dependencies for the parser. (Kafo::ParserError)
```

However, `--scenario foreman` works fine. Does something need to be added to each scenario?

ehelms:
@beav what version of kafo is installed?

beav:
@ehelms 
```bash
# rpm -qa | grep kafo | sort
rubygem-kafo-0.9.4-1.el7.noarch
rubygem-kafo_parsers-0.1.2-1.el7.noarch
rubygem-kafo_wizards-0.0.1-2.el7.noarch
```

ehelms:
@beav can you confirm that parser_cache exists at /usr/share/katello-installer and that /etc/foreman-installer/plugins/katello.yaml has the entry in it?

beav:
@ehelms 

```bash
# ls -ltrh /usr/share/katello-installer-base/parser_cache/katello.yaml 
-rw-r--r--. 1 root root 17K Aug 25 12:21 /usr/share/katello-installer-base/parser_cache/katello.yaml
```

The file `/etc/foreman-installer/scenarios.d/katello.yaml` did not have an entry for `parser_cache_path`, I added one to point to the file above but it did not seem to help.

ehelms:
@beav did you add https://github.com/Katello/katello-installer/pull/382/files#diff-63a01b9f79b464a7f20b06f0e1a3309eR12 ?

beav:
@ehelms I reset my vm and did the following:

```bash
rpm -Uvh 'http://koji.katello.org/koji/getfile?taskID=518316&name=katello-installer-base-3.2.0-2.nightly.git.0.4cbd175.el7.noarch.rpm' 'http://koji.katello.org/koji/getfile?taskID=518316&name=foreman-installer-katello-3.2.0-2.nightly.git.0.4cbd175.el7.noarch.rpm'
rpm -ivh 'http://koji.katello.org/koji/getfile?taskID=518316&name=foreman-installer-katello-devel-3.2.0-2.nightly.git.0.4cbd175.el7.noarch.rpm'
```

I then added the following to `/etc/foreman-installer/scenarios.d/katello-devel.yaml` and `katello.yaml`:
```yaml
:parser_cache_path:
  - /usr/share/foreman-installer/parser_cache/foreman.yaml
  - /usr/share/katello-installer-base/parser_cache/katello.yaml
```

This resulted in:
```bash
# foreman-installer --scenario katello --help
/usr/share/gems/gems/kafo-0.9.4/lib/kafo/configuration.rb:267:in `expand_path': no implicit conversion of Array into String (TypeError)
```

It looks like maybe I need a newer kafo, and something needs to be added to `katello.yaml`?

ehelms:
The `katello.yaml` change is here in this PR and the scratch build; however, depending on the order oyu installed things, if an existing installer package was on the system, it can result the config file not properly having the values it needs. This is a non-typical situation.

The kafo error is strange -- maybe Kafo is missing a change that is required for this.

beav:
@ehelms I can try to do a clean install without upgrading anything. How would it work during upgrades from say, 3.1 to 3.2?

ehelms:
Once I made this change [1] I got passing Puppet 3 server and capsule installations.

[1] https://github.com/theforeman/kafo/pull/172

beav:
@ehelms retesting now

beav:
@komidore64 and I took a look at this, it appears that this is causing the cache to not get picked up:

```ruby
        if !cache.is_a?(Hash) || cache[:version] != PARSER_CACHE_VERSION || !cache[:files].is_a?(Hash)
          logger.warn "Parser cache #{cache_paths[i]} is from a different version of Kafo, skipping setup"
          return nil
        end
```

The `katello.yaml` in the scratch build will be false for the first condition, but true for the second two.

komidore64:
the parser cache in the scratch build that @ehelms made was borked. i broke it when i fiddled it in the rakefile. updating the PR in a couple mins.

ehelms:
@beav @komidore64 new scratch build based on current state of PR:

http://koji.katello.org/koji/taskinfo?taskID=520706

beav:
@ehelms @komidore64 i need to perform a full runthrough before ack, but the new RPM worked OK for me :fireworks:

beav:
@ehelms @komidore64 new rpm works for me. I hit an issue but it was unrelated to this.

ACK

ehelms:
ACK - thanks @komidore64 !

PR #385, opened 2016-08-08
Fixes #16011 - handle both puppet 3 and 4 dirs when cleaning

beav:
Previously, the clear puppet environments flag would only clear the
puppet 3 dir. This patch iterates over both the puppet 3 and puppet 4
dirs, cleaning both if the flag is set.


ekohl:
ACK

beav:
@ekohl thanks for review/ack :koala:

PR #386, opened 2016-08-10
Fixes #16053: hooks for puppet 4 upgrade

beav:
This enables a `--upgrade-puppet` flag that will upgrade a Katello
installation from Puppet 3 to Puppet 4.

Most of this is based on
http://projects.theforeman.org/projects/foreman/wiki/Upgrading_from_Puppet_3_to_4
and https://github.com/beav/forklift/pull/1.

todo (feel free to add/subtract from this):

- [x] bail out before we start mucking with configs:
 - [x] exit early if puppet 4 packages are not available
 - [x] ensure directories we are copying from and to are readable/writable
- [x] test capsule

### DIFF
--- a/hooks/pre/30-upgrade.rb
+++ b/hooks/pre/30-upgrade.rb
@@ -94,6 +94,9 @@ def fail_and_exit(message)
 end
 
 if app_value(:upgrade)
+  fail_and_exit 'Concurrent use of --upgrade and --upgrade-puppet is not supported. '\
+                'Please run --upgrade first, then --upgrade-puppet.' if app_value(:upgrade_puppet)
+
   Kafo::Helpers.log_and_say :info, 'Upgrading...'
   katello = Kafo::Helpers.module_enabled?(@kafo, 'katello')
   capsule = @kafo.param('foreman_proxy_plugin_pulp', 'pulpnode_enabled').value
--- a/hooks/pre/31-upgrade-puppet.rb
+++ b/hooks/pre/31-upgrade-puppet.rb
@@ -1,0 +1,69 @@
+def puppet4_available?
+  success = []
+  success << Kafo::Helpers.execute('yum info puppet-agent')
+  success << Kafo::Helpers.execute('yum info puppetserver')
+  !success.include?(false)
+end
+
+def stop_services
+  Kafo::Helpers.execute('katello-service stop')
+end
+
+def upgrade_puppet_package
+  Kafo::Helpers.execute('yum remove -y puppet-server')
+  Kafo::Helpers.execute('yum install -y puppetserver')
+  Kafo::Helpers.execute('yum install -y puppet-agent')
+end
+
+def copy_data
+  success = []
+  success << Kafo::Helpers.execute('cp -rfp /etc/puppet/environments/* /etc/puppetlabs/code/environments') if File.directory?('/etc/puppet/environments')
+  success << Kafo::Helpers.execute('mv /var/lib/puppet/ssl /etc/puppetlabs/puppet') if File.directory?('/var/lib/puppet/ssl')
+  success << Kafo::Helpers.execute('mv /var/lib/puppet/foreman_cache_data /opt/puppetlabs/puppet/cache/') if File.directory?('/var/lib/puppet/foreman_cache_data')
+  !success.include?(false)
+end
+
+def upgrade_step(step)
+  noop = app_value(:noop) ? ' (noop)' : ''
+
+  Kafo::Helpers.log_and_say :info, "Upgrade Step: #{step}#{noop}..."
+  unless app_value(:noop)
+    status = send(step)
+    fail_and_exit "Upgrade step #{step} failed. Check logs for more information." unless status
+  end
+end
+
+def fail_and_exit(message)
+  Kafo::Helpers.log_and_say :error, message
+  kafo.class.exit 1
+end
+
+def reset_value(param)
+  unless app_value(:noop)
+    param.value = param.default unless param.nil?
+  end
+end
+
+if app_value(:upgrade_puppet)
+  katello = Kafo::Helpers.module_enabled?(@kafo, 'katello')
+  capsule = @kafo.param('foreman_proxy_plugin_pulp', 'pulpnode_enabled').value
+
+  fail_and_exit 'Puppet 3 to 4 upgrade is not currently supported for the chosen scenario.' unless katello || capsule
+
+  Kafo::Helpers.log_and_say :info, 'Upgrading puppet...'
+  fail_and_exit 'Unable to find Puppet 4 packages, is the repository enabled?' unless puppet4_available?
+
+  reset_value(param('foreman', 'puppet_home'))
+  reset_value(param('foreman', 'puppet_ssldir'))
+  reset_value(param('foreman_proxy', 'puppet_ssl_ca'))
+  reset_value(param('foreman_proxy', 'puppet_ssl_cert'))
+  reset_value(param('foreman_proxy', 'puppet_ssl_key'))
+  reset_value(param('foreman_proxy', 'puppetdir'))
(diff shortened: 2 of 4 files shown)


hooks/pre/31-upgrade-puppet.rb
```diff
@@ -0,0 +1,54 @@
...
+end
+
+def upgrade_puppet_pkg
+  Kafo::Helpers.execute('yum remove -y puppet-server')
+  Kafo::Helpers.execute('yum install -y puppetserver')
+  Kafo::Helpers.execute('yum install -y puppet-agent')
+end
+
+def populate_p4_environments
+  Kafo::Helpers.execute('cp -rfp /etc/puppet/environments/* /etc/puppetlabs/code/environments')
```
ekohl: Maybe include ```-a``` as well?

hooks/pre/31-upgrade-puppet.rb
```diff
@@ -0,0 +1,54 @@
...
+end
+
+def copy_ssl_certs
```
ekohl: This doesn't just copy the certs, also the (answer) cache. Maybe copy_data would be a better name?

ekohl:
And for the record, I like the idea of doing these more complex migrations via hooks. I do wonder if we can make them more generic so they work on Debian as well (or check if the OS supports the hook).

hooks/pre/31-upgrade-puppet.rb
```diff
@@ -1,20 +1,26 @@
-def stop_services                                                                                                                                                                                                                                     [0/1892]
+def puppet4_available?
+  success = []
+  success << Kafo::Helpers.execute('yum info puppet-agent')
+  success << Kafo::Helpers.execute('yum info puppetserver')
```
beav: ah cool, I was trying `yum search` instead of `yum info` and always getting zero exit code even if the pkg wasn't found. This change works as expected.

hooks/pre/31-upgrade-puppet.rb
```diff
@@ -1,20 +1,26 @@
...
 end
 
-def populate_p4_environments
-  Kafo::Helpers.execute('cp -rfp /etc/puppet/environments/* /etc/puppetlabs/code/environments')
-end
-
-def copy_ssl_certs
-  Kafo::Helpers.execute('mv /var/lib/puppet/ssl /etc/puppetlabs/puppet')
-  Kafo::Helpers.execute('mv /var/lib/puppet/foreman_cache_data /opt/puppetlabs/puppet/cache/')
+def copy_data
```
ekohl: I hate to be bikeshedding here, but given there's a move, should it just be ```migrate_files```? It also raises the question if we should move it all or copy it all instead of copy manifests but move certificates and caches.

ehelms:
I mentioned this in the docs PR -- but what about making the `--upgrade-puppet` and `--upgrade` mutually exclusive from one another?

hooks/pre/31-upgrade-puppet.rb
```diff
@@ -0,0 +1,65 @@
...
+
+def upgrade_puppet_package
+  Kafo::Helpers.execute('yum remove -y puppet-server')
+  Kafo::Helpers.execute('yum install -y puppetserver')
+  Kafo::Helpers.execute('yum install -y puppet-agent')
+end
+
+def copy_data
+  success = []
+  success << Kafo::Helpers.execute('cp -rfp /etc/puppet/environments/* /etc/puppetlabs/code/environments') if File.directory?('/etc/puppet/environments')
```
ehelms: As long as the next time a publish occurs, the new path is gotten from the smart proxy, we shouldn't have any issues here with respect to already published data because we tell Pulp the root value to use?

ehelms:
In theory, we are kicking this off in Puppet 3, removing puppet 3 while executing and continue on in memory of sorts? Should we do the puppet-agent package upgrade aspect as a post-install hook to upgrade the agent package last?

beav:
@ehelms i think it swaps over to puppet 4 before puppet gets fired off, the updates happen in the pre hook

beav:
I just did a rebase, these patches applied cleanly

hooks/pre/30-upgrade.rb
```diff
@@ -94,6 +94,8 @@ def fail_and_exit(message)
 end
 
 if app_value(:upgrade)
+  fail_and_exit 'Upgrading Katello and Puppet at the same time is not supported. Please upgrade Katello first, then upgrade Puppet' if app_value(:upgrade_puppet)
```
stbenjam: :+1:

hooks/pre/31-upgrade-puppet.rb
stbenjam: Doesn't seem to be, I tested a p3 -> p4 upgrade that had published content and it was working.

hooks/pre/31-upgrade-puppet.rb
```diff
@@ -0,0 +1,68 @@
...
+def reset_value(param)
+  param.value = param.default unless param.nil?
+end
+
+if app_value(:upgrade_puppet)
+  katello = Kafo::Helpers.module_enabled?(@kafo, 'katello')
+  capsule = @kafo.param('foreman_proxy_plugin_pulp', 'pulpnode_enabled').value
+
+  fail_and_exit 'Puppet 3 to 4 upgrade currently requires Katello or Katello capsule' unless katello || capsule
+  fail_and_exit 'Upgrading Katello and Puppet at the same time is not supported. Please upgrade Katello first, then upgrade Puppet' if app_value(:upgrade)
```
stbenjam: We can remove this one, as it's already handled in 30-upgrade.rb

hooks/pre/31-upgrade-puppet.rb
```diff
@@ -0,0 +1,68 @@
...
+  end
+end
+
+def fail_and_exit(message)
+  Kafo::Helpers.log_and_say :error, message
+  kafo.class.exit 1
+end
+
+def reset_value(param)
+  param.value = param.default unless param.nil?
```
stbenjam: Not 100% sure but I think we should check for app_value(:noop) here

beav:
@stbenjam any issue if i rebase/squash w/ next round of changes?

stbenjam:
@beav Nope, feel free

hooks/pre/30-upgrade.rb
ehelms: :+1:

beav:
PR updated to make two error strings a bit more agnostic :space_invader:

beav:
@ehelms this should work on capsule. I hit an error on Friday, but it appears to be resolved now with latest nightlies. The parser configs in `capsule.yaml` look correct as well, not sure what happened on Friday.

ehelms:
Probably a related but fixable as a separate issue, I see this output during my upgrade (and I assume it is due to the update to Puppet 4):

```
 Facter: error while resolving custom fact "mongodb_is_master": 757: unexpected token at '2016-09-06T12:19:47.235-0400 warning: Failed to connect to 127.0.0.1:27017, reason: errno:111 Connection refused
```

beav:
@ehelms i saw that as well, you should still get green success at the end. I don't remember offhand why that error pops up.

ehelms:
Tested P3 server and capsule upgrades without issue other than hitting (http://projects.theforeman.org/issues/16455) which is unrelated directly to this change. ACK

stbenjam:
@ehelms I fixed that error over here https://github.com/puppetlabs/puppetlabs-mongodb/pull/322, just need a new release of puppetlabs-mongodb.

ehelms:
@stbenjam I guess the whole team could ask them really nicely or we just point to git

stbenjam:
It is a harmless error, can we wait until we cut the first RC right before we decide to pin the installer to git? 

I just asked on #puppet-dev on freenode for a release.

PR #390, opened 2016-08-26
Fixes #16354 - Add rubocop

daviddavis:


hooks/pre/30-upgrade.rb
```diff
@@ -72,7 +72,7 @@ def fix_katello_settings_file
 
   return true unless settings.key?(:common)
 
-  settings = {:katello => settings[:common]}
+  settings = { :katello => settings[:common] }
```
ehelms: Is this normal? Seems strange to have a leading space.

ehelms:
Nice! I'd be thinking about this but of course you are the king of rubocop.

hooks/pre/30-upgrade.rb
daviddavis: That is odd but it seems that's [what rubocop does by default](https://git.io/v6x9L). I'm going to copy over the setting from katello: https://git.io/v6x9s

daviddavis:
This should be good to go 👑

ehelms:
ACK - thanks @daviddavis

PR #392, opened 2016-09-06
Refs #16354 - Hardcoding ruby version

daviddavis:
People using Ruby 2.3.0 are getting a different set of rubocop warnings/errors when they run rubocop locally. This change will prevent that. Using Ruby version 2.0 based on https://git.io/viZBa.


komidore64:
ACK

ehelms:
Thanks @daviddavis !

PR #393, opened 2016-09-06
fixes #16455 - parser cache for capsule-certs-generate

komidore64:


bin/capsule-certs-generate
```diff
@@ -45,11 +45,20 @@ CONFIG_CONTENT = <<EOF
 EOF
 
 config_file = temp_file('capsule-certs', CONFIG_CONTENT)
+
+# generate our parser cache and add it to the config file
+parser_cache = temp_file('capsule-certs.parser_cache', `kafo-export-params -c #{config_file.path} -f parsercache --no-parser-cache`)
```
ehelms: Can we do this on a production box? I assume it would require us to package puppet-strings which we were trying to avoid?

komidore64: are you asking if we **should**? because i know we can ;)

ehelms: Correct -- given our previous reasons we were avoiding packaging puppet-strings.

komidore64: Yeah, we're doing parser caches to avoid packaging puppet-strings.

komidore64: now i understand your roundabout question. yeah, this wouldn't work on a production machine because there would be no puppet strings to generate the parser cache. dunno why i didn't catch that. reworking.

ehelms: This is a tough one due to the inlined configuration. We didn't during the scenario work because capsule-certs-generate is unique in that it is not an installer scenario but uses kafo and our modules. So its more like a script than anything. We could:

Option 1:
 a) change this to lay down a config file
 b) have that separate from the bin file and not inlined anymore
 c) generate parser cache by pointing at the config file

Option 2: 
 a) see if there is some way to load the file via a script
 b) pass the contents of the CONFIG_CONTENT constant to kafo-export-params in the Rakefile

mbacovsky: I like Option 1/ more. But it would require not to mix the real scenarios with capsule-certs-generate configuration. How about to move the current content of `config` to `scenarios.d` and `capsule-certs-generate.yaml` keep in a `config`. Then `scenarios.d` would be installed to `/etc/foreman-installer/scenarios.d` and config to `/etc/foreman-installer-katello/` or better to `/usr/share/foreman-installer-katello/config/`.

komidore64:
@ehelms, @mbacovsky updated!

bin/capsule-certs-generate
```diff
@@ -3,53 +3,75 @@ require 'rubygems'
...
+                    :manifest_name: capsule
+                    :dir_name: certs
+                :order:
+                  - certs
+                  - capsule_certs
+              EOF
+             )
+  end
+
+  def initialize
```
ehelms: Nitpick - I generally think to look for this at the top since its a special function that tells you how a class is setup.

bin/capsule-certs-generate
```diff
@@ -3,53 +3,75 @@ require 'rubygems'
...
+    @answers_file = generate_answers_file
+    @config_file = generate_config_file(answers_file)
+  end
+
+  def temp_file(name, content)
+    f = Tempfile.open([name, '.yaml'])
+    f.write content
+    f.close
+    f
```
ehelms: Nitpick -- prefer readable variables over single characters

bin/capsule-certs-generate
```diff
@@ -3,53 +3,75 @@ require 'rubygems'
...
+                  deploy: false
+                  group: foreman
+                  org: Default_Organization
+                capsule_certs: true
+              EOF
+             )
+  end
+
+  # rubocop: disable Metrics/MethodLength
+  def generate_config_file(ans_f)
```
ehelms: Nitpick - readable (even if longer) variables help the reviewer and readers later of code to spend less time interpreting (e.g. ans_f vs. answer_file).

ehelms:
Some code nitpicks - going to test this now.

bin/capsule-certs-generate
komidore64: moved up top

bin/capsule-certs-generate
komidore64: changed to `file`

bin/capsule-certs-generate
komidore64: removed. we'll just use the `attr_reader` instead.

komidore64:
nitpicks addressed :)

komidore64:
updated again (sorry).

probably should ensure that `cleanup()` is run.

bin/capsule-certs-generate
```diff
@@ -3,53 +3,75 @@ require 'rubygems'
...
-
-Kafo::KafoConfigure.run
-
-config_file.unlink
-answers_file.unlink
-
-Kafo::KafoConfigure.exit_code == 2 ? exit(0) : exit(Kafo::KafoConfigure.exit_code)
+if __FILE__ == $0
+  gen = CapsuleCertsGenerate.new
+  CONFIG_FILE = gen.config_file
```
ehelms: I think this needs to be `gen.config_file.path`

bin/capsule-certs-generate
```diff
@@ -3,53 +3,78 @@ require 'rubygems'
...
-Kafo::KafoConfigure.run
-
-config_file.unlink
-answers_file.unlink
-
-Kafo::KafoConfigure.exit_code == 2 ? exit(0) : exit(Kafo::KafoConfigure.exit_code)
+if __FILE__ == $0
+  begin
+    gen = CapsuleCertsGenerate.new
+    CONFIG_FILE = gen.config_file
```
ehelms: Still needs to end in `.path`

bin/capsule-certs-generate
```diff
@@ -3,53 +3,78 @@ require 'rubygems'
...
+    Kafo::KafoConfigure.run
+  ensure
+    gen.cleanup
```
ehelms: :+1: to the ensure

ehelms:
I applied this on my test environment, but still getting:

```
[root@centos7-katello-p4-nightly vagrant]# capsule-certs-generate --capsule-fqdn centos7-capsule-p4-nightly.example.com --certs-tar /root/centos7-capsule-p4-nightly.example.com.tar.gz
/usr/share/gems/gems/kafo-0.9.5/lib/kafo/puppet_module.rb:69:in `parse': No Puppet module parser is installed and no cache of the file /usr/share/katello-installer-base/modules/certs/manifests/init.pp is available. Please check debug logs and install optional dependencies for the parser. (Kafo::ParserError)
        from /usr/share/gems/gems/kafo-0.9.5/lib/kafo/configuration.rb:88:in `block in modules'
        from /usr/share/gems/gems/kafo-0.9.5/lib/kafo/configuration.rb:88:in `map'
        from /usr/share/gems/gems/kafo-0.9.5/lib/kafo/configuration.rb:88:in `modules'
        from /usr/share/gems/gems/kafo-0.9.5/lib/kafo/configuration.rb:188:in `params'
        from /usr/share/gems/gems/kafo-0.9.5/lib/kafo/configuration.rb:198:in `preset_defaults_from_puppet'
        from /usr/share/gems/gems/kafo-0.9.5/lib/kafo/kafo_configure.rb:251:in `set_parameters'
        from /usr/share/gems/gems/kafo-0.9.5/lib/kafo/kafo_configure.rb:97:in `initialize'
        from /usr/share/gems/gems/clamp-1.0.0/lib/clamp/command.rb:133:in `new'
        from /usr/share/gems/gems/clamp-1.0.0/lib/clamp/command.rb:133:in `run'
        from /usr/share/gems/gems/kafo-0.9.5/lib/kafo/kafo_configure.rb:148:in `run'
        from /sbin/capsule-certs-generate:74:in `<main>'
```

Not sure if it doesn't like the cache I generated or something with kafo is occurring here.

ehelms:
Just https://github.com/Katello/katello-installer/pull/393#discussion_r77912818 needs handling for me, @mbacovsky if you have time tomorrow morning can you take a glance? We'd like to get things merged up before branching but I appreciate your thoughts.

bin/capsule-certs-generate
komidore64: doh!

komidore64: `.path` added

komidore64:
updated

mbacovsky:
@ehelms, @komidore64 Option 2/ makes the `capsule-certs-generate` even more special in my eyes but the code looks good. I appreciate the refactorings and the cleanup. +1 from me, I didn't have time to test it though.

ehelms:
@mbacovsky it's a special beast all around, one that I would love to re-design in general

ACK - thanks @komidore64 !

PR #394, opened 2016-09-07
fixes #16469 - exclude rubocop from checking _build and modules dirs

komidore64:


daviddavis:
ACK

ehelms:
Thanks @komidore64 !

PR #396, opened 2016-09-14
Locking puppet module versions for 3.2

johnpmitsch:


stbenjam:
I released some new versions yesterday to fix some problems that @beav found, so these aren't necessarily correct.  

I prefer we don't pin to exact versions. This is  good example why it's a bad idea.  Patch level versions are guaranteed compatible.  I know it's been done for the last few releases, and you can really do it however you want I guess. But if you pin to a range, you just need to run librarian to update the modules instead of updating the puppetfile every time.

This is how I think it should be done with a range:
https://github.com/Katello/katello-installer/blob/KATELLO-2.4/Puppetfile#L29-L41

stbenjam:
Also for release branches I think we still need to run the rake build task to actually commit the module contents, although not sure if anything changed recently

Puppetfile
```diff
@@ -1,14 +1,14 @@
...
-mod 'katello/certs',            :git => 'https://github.com/Katello/puppet-certs'
-mod 'katello/common',           :git => 'https://github.com/Katello/puppet-common'
-mod 'katello/crane',            :git => 'https://github.com/Katello/puppet-crane'
-mod 'katello/katello',          :git => 'https://github.com/Katello/puppet-katello'
-mod 'katello/pulp',             :git => 'https://github.com/Katello/puppet-pulp'
-mod 'katello/qpid',             :git => 'https://github.com/Katello/puppet-qpid'
-mod 'katello/service_wait',     :git => 'https://github.com/Katello/puppet-service_wait'
+mod 'katello/candlepin',        '1.1.0'
+mod 'katello/capsule',          '3.1.2'
+mod 'katello/certs',            '0.7.0'
```
stbenjam: 0.7.1

PR #397, opened 2016-09-14
Fixes #16561 - Don't copy modules dir to build dir

johnpmitsch:


Rakefile
```diff
@@ -22,12 +22,7 @@ file PARSER_CACHE_DIR => [BUILDDIR] do
 end
 
 file "#{BUILDDIR}/modules" do |t|
-  if Dir["modules/*"].empty?
-    sh "librarian-puppet install --verbose --path #{BUILDDIR}/modules"
-  else
-    mkdir BUILDDIR
-    cp_r "modules", "#{BUILDDIR}/modules"
-  end
+  sh "librarian-puppet install --verbose --path #{BUILDDIR}/modules"
```
ehelms: Grrrrreat

stbenjam: 6d7d8522e18c4cc363dc9b0dcb101d5f

ekohl:
I don't quite get this. The modules are frozen and imported into ```modules/```. Then shouldn't those modules be used instead?

A possible downside of this solution is that you can't build with local modules anymore. That means you can't build with local changes. Note I have no clue how often this is needed.

I'd also like to point to foreman-installer which uses a different approach: commit ```Puppetfile.lock``` to git with exact pins. That doesn't bloat up the repository and doesn't require this change.

ehelms:
@ekohl I believe this stems from if you have stale modules in `modules/` from testing or what have you, then it has the potential to re-use and copy those old `modules/` when generating a tarball. Which when releasing a build, nightly or for a release has the potential to lead to a bad build whereas this is trying to ensure for a build that always fresh modules are installed.

ekohl:
@ehelms I still don't know if that's a valid reason for removing this option. Maybe the release process needs cleanup step?

johnpmitsch:
@ehelms what do you want to do with this PR?

ehelms:
@johnpmitsch looks like it can be closed now based on the diff

PR #398, opened 2016-09-14
Fixes #16300 - add overcommit memory check

chris1984:
added a check to make sure vm.overcommit is not enabled, which caused some issues downstream

### DIFF
--- a/hooks/pre/18-vm_overcommit.rb
+++ b/hooks/pre/18-vm_overcommit.rb
@@ -1,0 +1,11 @@
+# Check to see if vm.overcommit is enabled and if so exit but also give the user an option to disable the check with the --disable-system-checks flag
+if app_value(:disable_system_checks)
+  logger.warn 'Skipping system checks.'
+else
+  # Grab the current status of vm.overcommit from sysctl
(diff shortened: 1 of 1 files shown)


hooks/pre/18-vm_overcommit.rb
```diff
@@ -0,0 +1,10 @@
+# Grab the current status of vm.overcommit from sysctl
+overcommit_check = `sysctl -a | grep "overcommit_memory" | awk '{print $3}'`.to_i
```
stbenjam: Can be refactored to just `sysctl -n vm.overcommit_memory`

chris1984: Learned something new today :+1:  will fix

chris1984:
@stbenjam hey when you get a chance can you check the updated commit? It fails on the build but i dont think its related to my code since i saw other builds failing with the same rake error.

chris1984:
@stbenjam when you get a chance can I get a merge? looks like the job built ok

stbenjam:
Thanks @chris1984 !

PR #399, opened 2016-09-20
bump kafo to 1.0.1

beav:
This won't build with less than 1.0.1


beav:
@ekohl thanks for the fast review :koala:

PR #400, opened 2016-09-21
Fixes #15781 - add ssl_cert_file to env check

jlsherrill:


checks/env_variables.rb
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env ruby
+
+variables = %w(http_proxy https_proxy ssl_cert_file)
+
+if variables.map{ |variable| ENV[variable] }.compact.any?
```
ekohl: rubocop wants a space between ```map``` and ```{```

checks/env_variables.rb
```diff
@@ -0,0 +1,8 @@
...
+  $stderr.puts "Please unset the following environment variables before running the installer: #{variables.join(', ')}"
+  exit 1
```
ekohl: And trailing whitespace

checks/env_variables.rb
jlsherrill: updated

checks/env_variables.rb
jlsherrill: updated!

jlsherrill:
Thanks for the review @ekohl !

PR #402, opened 2016-09-26
Fixes #13546 - Enable TFTP during install

chris1984:


chris1984:
References this one https://github.com/Katello/katello-installer/pull/377

stbenjam:
Thanks, with this change, tftp is on by default with the katello scenario.  Do we need this on capsules too? https://github.com/Katello/katello-installer/blob/master/config/capsule.migrations/160120212700-foreman-proxy.rb#L34

chris1984:
@stbenjam thanks for catching that, updated with the capsule migration as well.

stbenjam:
Thanks @chris1984!

PR #404, opened 2016-10-03
Fixes #16778 - Added check for /dev/log on installer

chris1984:


hooks/pre/19-devlog.rb
```diff
@@ -0,0 +1,11 @@
+# check to see if /dev/log directory exists
+
+dir_present = File.exist?('/dev/log')
```
ekohl: Maybe move this to the elsif statement so skipping the check really does skip everything related?

chris1984:
@ekohl fixed

hooks/pre/19-devlog.rb
```diff
@@ -0,0 +1,10 @@
+# check to see if /dev/log directory exists
+
+if app_value(:disable_system_checks)
+  logger.warn 'Skipping system checks.'
+  # if dir_present is true then give warning to user and exit otherwise move on withe install
```
sthirugn: s/withe/with the

hooks/pre/19-devlog.rb
```diff
@@ -0,0 +1,10 @@
...
+elsif
```
ekohl: Shouldn't this be on a single line?

chris1984:
@ekohl @sthirugn fixed the typo and added the rubocop fix with the true/false statement on the same line as the elsif.

jlsherrill:
Thanks @chris1984, merging!

PR #406, opened 2016-10-04
Fixes #16789 - add ansible installer migrations

iNecas:


config/katello.migrations/160919173201_smart_proxy_ansible.rb
```diff
@@ -0,0 +1,2 @@
+answers['foreman_proxy::plugin::ansible'] ||= false
```
ekohl: Any reason you're not including this in ```config/katello.migrations/160405121109_foreman_ansible.rb```?

config/katello.migrations/161004173201_foreman_ansible.rb
```diff
@@ -0,0 +1,5 @@
+answers['foreman::plugin::ansible'] ||= false
+scenario[:mapping]['foreman::plugin::ansible'] ||= {:dir_name => 'foreman', :manifest_name => 'plugin/ansible'}
```
ekohl: Not sure if it was this before, but this is a string and the foreman_proxy is a symbol. Is that a difference?

stbenjam:
Tested and works great, thanks @iNecas

PR #408, opened 2016-10-11
Fixes #16869 - only run post upgrade steps once

jlsherrill:


hooks/post/30-upgrade.rb
```diff
@@ -1,3 +1,7 @@
+require 'fileutils'
+
+STEP_DIRECTORY='/etc/foreman-installer/hooks/post/'
```
stbenjam: Can we use a directory out of `@kafo.config.app`? Just in case we stop using this dir.

Or, what do you think about having a hidden directory/file like we do for applied migrations? Maybe right inside the hook directory itself, we could have a .ran file or something.

hooks/post/30-upgrade.rb
```diff
@@ -83,7 +104,7 @@ def fail_and_exit(message)
     upgrade_step :restart_services
```
stbenjam: Isn't this an always run?

hooks/post/30-upgrade.rb
stbenjam: I guess it's fine since you do mkpath, but I think I'd prefer if it was a more obvious name

hooks/post/30-upgrade.rb
jlsherrill: oops!  good catch

stbenjam:
One more thought, can we add an installer option to run everything? I'd guess that'd be easier to tell a user, in case we need them to run the tasks again

ehelms:
:+1: to the general idea, this process could use some standardization similar to migrations but i think having a solution for current makes sense.

hooks/post/30-upgrade.rb
jlsherrill: I didn't see anything in @kafo.config.app that was useful.  Just references to:

/etc/foreman-installer/scenarios.d/katello-answers.yaml
and
/etc/foreman-installer/scenarios.d

i could parse /etc/foreman-installer/scenarios.d and go up one or something, but that might be just as risky as hardcoding.

stbenjam: That's fine then, just making it more obvious what it is should be good enough.  e.g.,  `applied_hooks` like we talked about on IRC

jlsherrill: updated

jlsherrill:
Added option to force  --force-upgrade-steps

PR #410, opened 2016-10-11
Fixes #16864 - Remove gutterball db on upgrade

chris1984:


stbenjam:
Ruby code itself looks great :+1: 

Could we do this in the already existent post hook or does it need done in `pre`? https://github.com/Katello/katello-installer/blob/master/hooks/post/30-upgrade.rb#L55

chris1984:
@stbenjam Fixed the merge conflict, thanks for reviewing. It needs to be in pre so we can remove the database before tomcat starts up otherwise we get the error that the database has active connections.

stbenjam:
@chris1984 No problem, thanks! ACK

PR #411, opened 2016-10-13
fixes #17201 - sync migrations and answers file

stbenjam:


iNecas:
I guess we need to pull in a new puppet-foreman_proxy as well, which explains, why it's a good idea to actually have the migrations and scenario file

stbenjam:
I think failures are unrelated - but it looks like when we build the parser cache for Katello, we're building it only from released versions:

```
[Librarian] Resolving theforeman-foreman_proxy (< 5.0.0, >= 4.0.0) <https://forgeapi.puppetlabs.com>
```

Since we don't have them in our puppet file pointing to git.

I think maybe we could change the Rake task to merge the two Puppetfile's.  We probably should grab it from the foreman-installer branch that corresponds to Katello branch (develop->master, 1.13-stable, KATELLO-3.2, etc)?

@komidore64 @ehelms What do you think?

stbenjam:
Added a rake task to load Kafo and migrate the answers. `rake config:migrate`.

Rakefile
```diff
@@ -22,6 +26,9 @@ file PARSER_CACHE_DIR => [BUILDDIR] do
 end
 
 file "#{BUILDDIR}/modules" do |t|
+  # Append Foreman's Puppetfile to ours, so we use the right version of the puppet modules for the cache
+  File.open('Puppetfile', 'a') { |f| f.write(open("https://raw.githubusercontent.com/theforeman/foreman-installer/#{FOREMAN_BRANCH}/Puppetfile").read) }
```
ehelms: Does this then generate a katello parser cache with all the modules in it?

Rakefile
```diff
@@ -113,6 +119,25 @@ rescue
   puts 'Rubocop not loaded'
 end
 
+namespace :config do
```
ehelms: The idea of this is to occasionally 'sync' the files in git with the migrations? Sort of like keeping a database schema in git for your Rails application?

Rakefile
stbenjam: The foreman modules already get installed because they're dependencies, but it's wrong the versions because they'll pulled from puppet forge

Rakefile
stbenjam: It should *always* be done. If you look at the original test failure:

```
mkdir /home/travis/build/Katello/katello-installer/_build/parser_cache

kafo-export-params -c config/katello.yaml -f parsercache --no-parser-cache -o /home/travis/build/Katello/katello-installer/_build/parser_cache/katello.yaml

D, [2016-10-27T14:52:16.819596 #3276] DEBUG -- : Using Puppet module parser KafoParsers::PuppetModuleParser

Manifest foreman_proxy/manifests/plugin/ansible.pp was not found in /home/travis/build/Katello/katello-installer/_build/modules, /home/travis/build/Katello/katello-installer/modules
```

Kafo is using the configs to generate the parser cache, but it doesn't appear to be running the migrations (maybe that's a kafo bug? I dunno), but it can't generate the cache against our config because the ansible module isn't in the version of modules we're pulling.

ehelms: As in every PR should have run this? That seems hard to enforce but we can give it a try. I still think we ought to consider treating the migrations and answer files like Rails does where the schema is the result of applying the migrations and one is not stored in the git repo.

ehelms:
@stbenjam want to add a Redmine issue to the commit?

stbenjam:
Updated

Rakefile
stbenjam: I think the Rails use case and the Kafo use case are much different, it's very difficult for me to look through a dozen migrations + the answer file to figure out what our canonical value is for a new install

ekohl:
I'm wondering if we can create a Travis check based on this. The config files (```config/*```) should always be fully migrated. That implies that after running the migrations against our configs there should be no difference (```git status --short config/```?). Would that make sense?

stbenjam:
Yea, I think so - I'll add some tests on it

PR #414, opened 2016-10-19
Fixes #16972 - Adding post hook to create ssl.conf file

chris1984:
Workaround until mod_ssl packaging is corrected.

### DIFF
--- a/hooks/post/11-create_sslconf.rb
+++ b/hooks/post/11-create_sslconf.rb
@@ -1,0 +1,7 @@
+# post hook to create an ssl.conf file to work-around this issue http://projects.theforeman.org/issues/16972 until mod_ssl packging is fixed.
+
+if File.file?('/etc/httpd/conf.d/ssl.conf')
+  logger.info 'ssl.conf is already present, skipping'
+else
+  File.open('/etc/httpd/conf.d/ssl.conf', 'w') { |file| file.write("# This file is managed by the foreman-installer, do not alter.") }
+end


hooks/post/11-create_sslconf.rb
```diff
@@ -0,0 +1,3 @@
+# post hook to create an ssl.conf file to work-around this issue http://projects.theforeman.org/issues/16972 until mod_ssl packging is fixed.
+
+File.open('/etc/httpd/conf.d/ssl.conf', 'w') { |file| file.write("# This file is managed by Puppet, do not alter.") }
```
jlsherrill: otherwise looks good!

stbenjam:
Needs a check if the file exists first, please.  It will get created on el6 just not el7.

Also, if anything ever changes and the file ends up getting created somehow during the installation phase, the hook will wipe out the contents of the file. :skull:

jlsherrill:
oh right, for el6 arggggggggggggggggggg

hooks/post/11-create_sslconf.rb
```diff
@@ -0,0 +1,7 @@
+# post hook to create an ssl.conf file to work-around this issue http://projects.theforeman.org/issues/16972 until mod_ssl packging is fixed.
+
+if File.file?('/etc/httpd/conf.d/ssl.conf') == false
+  File.open('/etc/httpd/conf.d/ssl.conf', 'w') { |file| file.write("# This file is managed by the foreman-installer, do not alter.") }
+else
+  logger.info 'ssl.conf is already present, skipping'
+end
```
stbenjam: Looks fine, a few ruby style things though.

Generally I think "if false + else" isn't a good code practice - harder to read. It's better to swap the conditions (although this is really a matter of opinion):

```ruby
if File.file?('/etc/httpd/conf.d/ssl.conf')
  logger.info 'ssl.conf is already present, skipping'
else
  File.open('/etc/httpd/conf.d/ssl.conf', 'w') { |file| file.write("# This file is managed by the foreman-installer, do not alter.") }
end
```
But at the very least, don't compare with false. Some methods won't return an actual false object but rather something else that is falsy like "nil".  It's very common in Rails itself that some things return nil instead of false.

An example:

```ruby
a = nil

if !a
  puts "this will print"
end

if a == false
  puts "this won't"
end
```

File.file? does return false, but it's a bad habit.  IMHO prefer "!File.file('foo')" instead, or avoid the negation altogether by swapping the conditions like the first suggestion

stbenjam:
Looks ok to me, is this urgent or can it wait a day or two to see what happens to the puppetlabs PR?

stbenjam:
There's been no movement on the puppetlabs PR so will merge this for now.

PR #417, opened 2016-10-25
fixes #17074 - set param value to nil for puppet upgrade

stbenjam:
"nil" will reset the parameter to the puppet default, param.default
contains the current value as it's the "default" from the
answers file if there's one specified (it's poorly named)

After upgrade, puppetca settings are now correct:

``` yaml
# cat /etc/foreman-proxy/settings.d/puppetca.yml 
---
# PuppetCA management
:enabled: https
:ssldir: /etc/puppetlabs/puppet/ssl
:puppetdir: /etc/puppetlabs/puppet
```

### DIFF
--- a/hooks/pre/31-upgrade-puppet.rb
+++ b/hooks/pre/31-upgrade-puppet.rb
@@ -40,7 +40,7 @@ end
 
 def reset_value(param)
   unless app_value(:noop)
-    param.value = param.default unless param.nil?
+    param.value = nil unless param.nil?
   end
 end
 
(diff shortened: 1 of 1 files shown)


stbenjam:
@beav if you have time to take a look, this should fix the problem QE found

beav:
@stbenjam looks good code-wise, will test in a few

stbenjam:
This is green now

beav:
ACK

sorry for the delay on testing this

stbenjam:
No problem, thanks @beav!

PR #419, opened 2016-10-28
Fixes #17126: remove_gutterball step is a post upgrade action

ehelms:


stbenjam:
That's not true, the database part has to be done in pre:

https://github.com/Katello/katello-installer/blob/5d34d4fb886850d8152558dae07f8d376ca41d6a/hooks/pre/30-upgrade.rb#L40

See the discussion in this PR https://github.com/Katello/katello-installer/pull/410#issuecomment-253027538

stbenjam:
Although not sure why you're getting that error, the method should be there

PR #422, opened 2016-11-02
Fixes #16864 - remove gutterball webapp directory

chris1984:
Still seeing web app dir after upgrades:

For upgrades from 6.1, that directory remains undeleted, so tomcat still deploys gutterball app despite no gutterball RPM present, no gutterball DB. And that still subscribes to the $(fqdn):event qpid queue and still causes https://access.redhat.com/solutions/2608341 - confirmed in field already.

Switching to an rm -rfv to make sure we remove the dir 100%

### DIFF
--- a/hooks/post/30-upgrade.rb
+++ b/hooks/post/30-upgrade.rb
@@ -62,7 +62,7 @@ def remove_gutterball
 
   ['tomcat', 'tomcat6'].each do |t|
     gutterball_dir = "/var/lib/#{t}/webapps/gutterball"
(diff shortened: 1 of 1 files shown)


ekohl:
Build failure is unrelated. Not quite sure why that is

stbenjam:
https://github.com/Katello/katello-installer/pull/411 fixes it, we're pulling Foreman modules from the Forge, and our modules from nightly - they're not compatible.  We have to build the parser cache by merging our and Foreman's puppetfile.

stbenjam:
Thanks @chris1984!

PR #423, opened 2016-11-04
Fixes #17214 - add more info about capsule tar

chris1984:
added location on capsule fqdn where to scp the tar file to, since downstream customers are putting it in different spots and causing issues when the installer runs.

### DIFF
--- a/hooks/post/10-post_install.rb
+++ b/hooks/post/10-post_install.rb
@@ -35,9 +35,9 @@ if [0, 2].include?(@kafo.exit_code)
 
     if Kafo::Helpers.module_enabled?(@kafo, 'katello')
       say <<MSG
-  * To install additional capsule on separate machine continue by running:
+  * To install an additional capsule on a separate machine run the following command:
 
-      capsule-certs-generate --capsule-fqdn "<%= color('$CAPSULE', :info) %>" --certs-tar "<%= color('~/$CAPSULE-certs.tar', :info) %>"
+      capsule-certs-generate --capsule-fqdn "<%= color('$CAPSULE', :info) %>" --certs-tar "<%= color('/root/$CAPSULE-certs.tar', :info) %>"
 
 MSG
     end
@@ -61,7 +61,7 @@ MSG
   Once this is completed run the steps below to start the smartproxy installation:
 
   1. Ensure that the foreman-installer-katello package is installed on the system.
-  2. Copy <%= color("#{certs_tar}", :info) %> to the system <%= color("#{capsule_fqdn}", :info) %>
+  2. Copy the following file <%= color("#{certs_tar}", :info) %> to the system <%= color("#{capsule_fqdn}", :info) %> at the following location <%= color("#{File.join('/root', File.basename(certs_tar))}", :info) %>
   3. Run the following commands on the capsule (possibly with the customized
      parameters, see <%= color("#{installer_name} --scenario capsule --help", :info) %> and
      documentation for more info on setting up additional services):


stbenjam:
>  added location on capsule fqdn where to scp the tar file to, since downstream customers are putting it in different spots and causing issues when the installer runs.


What issues are users running into? The certs tar should be able to be stored anywhere, the path you pass to the installer is absolute.

chris1984:
@stbenjam The email I got from the sat 6 docs writer stated that users are running the capsule-certs-generate command on the satellite and passing a dir then just scping the tar to the capsule somewhere different and coping the installer command that the capsule-certs-generate gives and then they get an error. This was just a tip to make sure users are actually putting in in the path so they can copy and paste the installer command without error.

hooks/post/10-post_install.rb
```diff
@@ -37,7 +37,7 @@ def new_install?
       say <<MSG
   * To install additional capsule on separate machine continue by running:
 
-      capsule-certs-generate --capsule-fqdn "<%= color('$CAPSULE', :info) %>" --certs-tar "<%= color('~/$CAPSULE-certs.tar', :info) %>"
+      capsule-certs-generate --capsule-fqdn "<%= color('$CAPSULE', :info) %>" --certs-tar "<%= File.join('/root', File.basename('certs.tar.gz')) %>"
```
stbenjam: File.basename on a string isn't going to get you anything, do it on `certs_tar` variable which comes from the puppet parameters, you might have to move some things around, since it's defined below this line

stbenjam: Actually, this is just the help, just leave the original suggestion, or make it a hard coded path, although the above comment applies to the other reference

hooks/post/10-post_install.rb
```diff
@@ -61,7 +61,7 @@ def new_install?
   Once this is completed run the steps below to start the smartproxy installation:
 
   1. Ensure that the foreman-installer-katello package is installed on the system.
-  2. Copy <%= color("#{certs_tar}", :info) %> to the system <%= color("#{capsule_fqdn}", :info) %>
+  2. Copy the following file <%= color("#{certs_tar}", :info) %> to the system <%= color("#{capsule_fqdn}", :info) %> at the following location <%= File.join('/root', File.basename('certs.tar.gz')) %>
```
stbenjam: Why File.basename on  a string?

chris1984:
Looks like this with new changes:

http://imgur.com/a/RuuDv

hooks/post/10-post_install.rb
```diff
@@ -61,7 +61,7 @@ def new_install?
   Once this is completed run the steps below to start the smartproxy installation:
 
   1. Ensure that the foreman-installer-katello package is installed on the system.
-  2. Copy <%= color("#{certs_tar}", :info) %> to the system <%= color("#{capsule_fqdn}", :info) %>
+  2. Copy the following file <%= color("#{certs_tar}", :info) %> to the system <%= color("#{capsule_fqdn}", :info) %> at the following location <%= color("#{certs_tar}", :info) %>
```
stbenjam: It looks better, but the problem is the user could run:

```
capsule-certs-generate --capsule-fqdn "foo.example.com" --certs-tar "certs.tar.gz"
 ```

Or worse:

```
capsule-certs-generate --capsule-fqdn "foo.example.com" --certs-tar "/home/joe/certs.tar.gz"
 ```

In both cases certs_tar will just have exactly what the user gave us, so it's not helping in suggesting a place when running capsule-certs-generate.

I think you should do this:

```
2. Copy the following file <%= color("#{certs_tar}", :info) %> to the system <%= color("#{capsule_fqdn}", :info) %> at the following location <%= color("#{File.join("/root", File.basename(certs_tar))}", :info) %> 
```

The first certs_tar is what the user gave us, but the suggested copy location would end up being /root/certs.tar.gz as the suggested destination.

chris1984:
@stbenjam I agree there is no way to fully fool proof from errors happening but anything we can do to help is much appreciated. I have made the following changes and tested them and it looks good with the installer on a scenario run telling them to use /root/certs.tar.gz and then if they do use that path then telling them to put it in /root/certs_tar variable :info

hooks/post/10-post_install.rb
```diff
@@ -35,9 +35,9 @@ def new_install?
 
     if Kafo::Helpers.module_enabled?(@kafo, 'katello')
       say <<MSG
-  * To install additional capsule on separate machine continue by running:
+  * To install an additional capsule on a separate machine run the following command:
 
-      capsule-certs-generate --capsule-fqdn "<%= color('$CAPSULE', :info) %>" --certs-tar "<%= color('~/$CAPSULE-certs.tar', :info) %>"
+      capsule-certs-generate --capsule-fqdn "<%= color('$CAPSULE', :info) %>" --certs-tar "/root/certs.tar.gz"
```
stbenjam: I don't think this should be changed, except maybe giving a path. Running the command you're giving now, they'll overwrite any other certs tarballs they've created.  It should still have $CAPSULE in the name.

chris1984:
@stbenjam thanks again for the review, I sent in the requested changes, when you get a chance can you review again.

Thanks

hooks/post/10-post_install.rb
```diff
@@ -35,9 +35,9 @@ def new_install?
 
     if Kafo::Helpers.module_enabled?(@kafo, 'katello')
       say <<MSG
-  * To install additional capsule on separate machine continue by running:
+  * To install an additional capsule on a separate machine run the following command:
 
-      capsule-certs-generate --capsule-fqdn "<%= color('$CAPSULE', :info) %>" --certs-tar "<%= color('~/$CAPSULE-certs.tar', :info) %>"
+      capsule-certs-generate --capsule-fqdn "<%= color('$CAPSULE', :info) %>" --certs-tar "<%= color('/root/$CAPSULE-certs.tar.gz', :info) %>"
```
stbenjam: It's not a gz

chris1984:
@stbenjam sorry about that, i have always used .gz so didnt know it was an issue. Back to tar now.

stbenjam:
Thanks, looks good

PR #424, opened 2016-11-04
Fixes #16252 - add check to reset hook

chris1984:


hooks/pre/10-reset_feature.rb
```diff
@@ -18,6 +18,12 @@ def foreman_installed?
   $?.success?
 end
 
+response = ask('Are you sure you want to continue? [y/n]')
+unless response == 'y'
+  $stderr.puts '** cancelled **'
+  kafo.class.exit(1)
+end
```
stbenjam: This gets asked on every invocation of the installer even if you didn't specify `--reset`.

It's also in a bit of an awkward place since it's in the middle of a bunch of method defintions.  It should go to the bottom.

hooks/pre/10-reset_feature.rb
```diff
@@ -57,4 +63,11 @@ def reset_pulp
   Kafo::Helpers.execute(commands)
 end
 
-reset if app_value(:reset) && !app_value(:noop)
+if app_value(:reset) && !app_value(:noop)
+  response = ask('Are you sure you want to continue? [y/n]')
+  if response == 'n' || response == 'N'
+    $stderr.puts '** cancelled **'
+    kafo.class.exit(1)
+  elsif reset
```
ehelms: This look to be unneeded

hooks/pre/10-reset_feature.rb
```diff
@@ -57,4 +63,11 @@ def reset_pulp
   Kafo::Helpers.execute(commands)
 end
 
-reset if app_value(:reset) && !app_value(:noop)
+if app_value(:reset) && !app_value(:noop)
+  response = ask('Are you sure you want to continue? [y/n]')
+  if response == 'n' || response == 'N'
```
ehelms: You could optionally do `if response.lower == 'n'`

hooks/pre/10-reset_feature.rb
```diff
@@ -57,4 +63,11 @@ def reset_pulp
   Kafo::Helpers.execute(commands)
 end
 
-reset if app_value(:reset) && !app_value(:noop)
+if app_value(:reset) && !app_value(:noop)
+  response = ask('Are you sure you want to continue? [y/n]')
```
ehelms: Mind putting a bit more information in here for context for the user? Perhaps warn them what reset will do and ask if they wish to continue?

hooks/pre/10-reset_feature.rb
stbenjam: I think it was meant to be: 

```ruby
else
  reset
```

We still need the reset call

stbenjam: But, the logic should be reversed, only 'Y/y' should trigger the reset.  The way it's written now if a user types anything but N or n, it would reset

chris1984:
Talked to David Davis and he said it was ok to add that line to Rubocop since it is not letting me use .downcase

stbenjam:
The code is still here that asks every time

chris1984:
[root@rhel7 pre]# cat 10-reset_feature.rb 
if app_value(:reset) && !app_value(:noop)
  response = ask('Are you sure you want to continue? This will drop the databases, reset all configurations that you have made and bring the Katello server back to a fresh install.[y/n]')
  if response.downcase != 'y'
    $stderr.puts '** cancelled **'
    kafo.class.exit(1)
  else
    reset
  end
end

Tested on downstream latest snap 6.3 (since nightly is not working) and it works with just running the installer. Also updated the warning message about what the reset will do.

[root@rhel7 pre]# satellite-installer --scenario satellite -v --disable-system-checks
[ INFO 2016-11-22 17:48:03 verbose] Executing hooks in group pre_migrations
[ INFO 2016-11-22 17:48:03 verbose] All hooks in group pre_migrations finished
[ INFO 2016-11-22 17:48:03 verbose] Executing hooks in group boot

hooks/pre/10-reset_feature.rb
stbenjam: You still have this code block in your commit

stbenjam:
I don't know what you're testing, but it's not the code in this PR.  You have 2 places you're asking the answer for a response, just look at the files tab here:
 https://github.com/Katello/katello-installer/pull/424/files

chris1984:
@stbenjam Sorry about that I was testing the reset within the if statment at the bottom and forgot to remove it from above where it was first submitted. I have removed it from there with the latest commit.

stbenjam:
Thanks, looks good. ACK

FWIW when testing PR's, I don't copy and paste changes (usually), you can do a wget straight from github of a file, if you click the "View" link and then "Raw" you can get the url. 


E.g. https://raw.githubusercontent.com/chris1984/katello-installer/e51c5786387ad020b0350c66ce70e851469da353/hooks/pre/10-reset_feature.rb


At the very least review the diff in the Files tab to make sure it looks like you expect.

stbenjam:
@ehelms Can you make sure your requests were addressed

hooks/pre/10-reset_feature.rb
```diff
@@ -57,4 +57,12 @@ def reset_pulp
   Kafo::Helpers.execute(commands)
 end
 
-reset if app_value(:reset) && !app_value(:noop)
+if app_value(:reset) && !app_value(:noop)
+  response = ask('Are you sure you want to continue? This will drop the databases, reset all configurations that you have made and bring the Katello server back to a fresh install.[y/n]')
```
ehelms: Can we drop the 'Katello' in the middle there? I think server should be sufficient.

ehelms:
Thanks @chris1984 !

PR #425, opened 2016-11-04
refs #17201 - add rspec tests for migrations

stbenjam:


spec/migration_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    context "on #{scenario}" do
+      let(:answers) { File.expand_path(File.join(CONFIG_DIR, "#{scenario}-answers.yaml")) }
+      let(:config) { File.expand_path(File.join(CONFIG_DIR, "#{scenario}.yaml")) }
+
+      before(:each) do
+        ENV['INSTALLER_SCENARIO'] = scenario
+      end
+
+      after(:each) do
+        `git reset HEAD --hard`
```
ekohl: This may surprise developers who run tests. Maybe assert ```git status --short config/``` is empty?

stbenjam: Hm, good point. I'll just fix it so the test migrations don't dump their answers back into the git repo, which is the only reason I have the reset.

stbenjam:
@ekohl Updated

spec/migration_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+          :answers    => YAML.load_file(File.expand_path(File.join(CONFIG_DIR, "#{scenario_name}-answers.yaml"))),
+          :config     => YAML.load_file(File.expand_path(File.join(CONFIG_DIR, "#{scenario_name}.yaml"))),
+          :migrations => File.expand_path(File.join(CONFIG_DIR, "#{scenario_name}.migrations"))
+        }
+      end
+
+      let(:migrator) { Kafo::Migrations.new(scenario[:migrations]).run(scenario[:config].dup, scenario[:answers].dup) }
+
+      it 'does not change scenario config' do
+        after, _ = migrator
```
ekohl: rubocop doesn't like this

stbenjam:
Thanks!

ekohl:
Thanks! This also adds tests to all our migrations which I believe we didn't have at first. That's a cool addition.

stbenjam:
Indeed, maybe it's useful to foreman-installer? https://github.com/theforeman/foreman-installer/pull/208

PR #426, opened 2016-11-08
fixes #17269 - pin devel to 3.2-compat branch

stbenjam:


beav:
I need to test this with another 3.2 change to test it fully, but this appears to work great for me.

beav:
APJ

stbenjam:
Fixed the rubocop issue, it's green.  Thanks @beav!

PR #429, opened 2016-11-08
Fixes #17124 - install capsule parser cache

beav:
There were some fixes in master for capsule parser cache, this commit
is a smaller version to just pull what's needed to build the capsule
parser cache for puppet 4.

### DIFF
--- a/config/capsule.yaml
+++ b/config/capsule.yaml
@@ -8,13 +8,47 @@
   :answer_file: "./config/capsule-answers.yaml"
   :installer_dir: "."
   :default_values_dir: "/tmp"
-  :module_dirs: ["/usr/share/foreman/modules", "./modules"]
+  :module_dirs: ["./_build/modules", "../katello-installer/modules"]
   :parser_cache_path:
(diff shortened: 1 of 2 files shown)


beav:
this depends on https://github.com/Katello/katello-installer/pull/426

stbenjam:
#426 merged, restarted tests

config/capsule.yaml
```diff
@@ -8,13 +8,133 @@
...
   :colors: true
   :order:
     - certs
     - capsule
+  :mapping:
+    :foreman::cli:
+      :dir_name: foreman
+      :manifest_name: cli
+      :params_name: cli/params
+    :foreman::plugin::bootdisk:
```
stbenjam: Hm, these shouldn't be here in the capsule scenario, these are all foreman plugins. What added them?

stbenjam: I think only the foreman_proxy plugins should be here

beav:
updated to remove extraneous declarations in `capsule.yaml`

stbenjam:
ACK

beav:
@stbenjam thanks!

PR #435, opened 2016-11-17
Fixes #17331 - supress el6 qpid-stat traceback

chris1984:
Fixes this issue that is related to qpid, its not every upgrade or time either. This happens after an upgrade with a manual qpid-stat from the cli too. The queue gets removed just it gives a nasty traceback to the user during the upgrade process. I talked with @mccun934 and he is fine with just suppressing the error. Also added '' around my grep.

Traceback that appears during upgrade is at the bottom of the paste below:

https://paste.fedoraproject.org/483909/40169014/

### DIFF
--- a/hooks/post/30-upgrade.rb
+++ b/hooks/post/30-upgrade.rb
@@ -69,7 +69,7 @@ end
 def remove_event_queue
   queue_present = `qpid-stat -q --ssl-certificate=/etc/pki/katello/qpid_client_striped.crt -b amqps://localhost:5671 | grep :event | wc -l`.chomp.to_i
   if queue_present > 0
-    Kafo::Helpers.execute('qpid-config --ssl-certificate=/etc/pki/katello/qpid_client_striped.crt -b amqps://localhost:5671 del queue $(hostname -f):event --force')
+    Kafo::Helpers.execute('qpid-config --ssl-certificate=/etc/pki/katello/qpid_client_striped.crt -b amqps://localhost:5671 del queue $(hostname -f):event --force > /dev/null 2>&1')
   else
     logger.info 'Event queue is already removed, skipping'
   end


hooks/post/30-upgrade.rb
```diff
@@ -67,9 +67,9 @@ def remove_gutterball
 end
 
 def remove_event_queue
-  queue_present = `qpid-stat -q --ssl-certificate=/etc/pki/katello/qpid_client_striped.crt -b amqps://localhost:5671 | grep :event | wc -l`.chomp.to_i
+  queue_present = `qpid-stat -q --ssl-certificate=/etc/pki/katello/qpid_client_striped.crt -b amqps://localhost:5671 | grep ':event' | wc -l`.chomp.to_i
```
stbenjam: The bug says '' causes the error but that doesn't seem to be true, there's no special characters in `:event` that would be need to be treated as a string literal

stbenjam:
Do we know why it happens? I'm fine with suppressing the output, but I don't see this error on my own el6 box.

chris1984:
@stbenjam ill update the bz and change it to a qpid-bug I have asked the qpid-team in #messaging and they have not seen it but are going to look into it, ill file a separate qpid bug for them to look at and adjust this one to reflect the PR. I dont know enough about qpid or python to see whats causing it. 2 users have hit it downstream and i was out of 3 tests able to get it to reproduce 2 times. So it is random. With Justin's PR skipping steps that dont need to be run again hopefully by downstream 6.3 this will never be an issue.

chris1984:
@stbenjam heard back from the qpid quys:

trace typically comes from less-than-clean shutdown
the random part isn't too surprising.  it depends on the order in which the python interpreter tears down threads

stbenjam:
Yea suppressing is fine, I would prefer the first change though (the ':event') thingy was removed, so someone in the future isn't confused about what fixed the bug.

chris1984:
@stbenjam Thanks for reviewing, new changes are committed.

stbenjam:
Thanks, ACK

PR #436, opened 2016-11-21
fixes #17415 - remove unmanaged httpd conf files

stbenjam:
Like mod_ssl did, pulp package now places a pulp.conf that has a duplicate directive.

### DIFF
--- a/hooks/post/29-create_package_httpd_conf.rb
+++ b/hooks/post/29-create_package_httpd_conf.rb
@@ -1,0 +1,22 @@
+# Some packages like mod_ssl and pulp place configuration files in
+# /etc/httpd/conf.d that contain duplicates declarations of directives
+# that the puppetlabs-apache module configured elsewhere.
+#
+# * pulp places a pulp.conf that contains a duplicate WSGI 'pulp'
+#   daemon that 05-pulps-https.conf contains.
+#
+# * mod_ssl places a 'Listen 443' directive in ssl.conf that ports.conf already
+#   has.
+#
+# This hook creates placeholders files so the package does not put them
+# in place anymore.
+#
+%w(ssl.conf pulp.conf).each do |file|
+  if File.file?(File.join("/etc/httpd/conf.d/", file))
+    File.open(File.join('/etc/httpd/conf.d', file), 'w') do |f|
+      f.write("# This file is managed by the foreman-installer, do not alter.")
+    end
+  else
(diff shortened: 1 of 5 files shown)


hooks/post/29-create_package_httpd_conf.rb
```diff
@@ -0,0 +1,23 @@
...
+%w(ssl.conf pulp.conf).each do |file|
+  if File.file?(File.join("/etc/httpd/conf.d/", file))
+    File.open(File.join('/etc/httpd/conf.d', file), 'w') do |f|
+      f.write("# This file is managed by the foreman-installer, do not alter.")
+    end
+  else
+    logger.info 'ssl.conf is already present, skipping'
+  end
+end
+
```
ekohl: Nit: empty line

hooks/pre/30-upgrade.rb
```diff
@@ -74,12 +74,6 @@ def remove_nodes_distributors
   Kafo::Helpers.execute("mongo pulp_database --eval  'db.repo_distributors.remove({'distributor_type_id': \"nodes_http_distributor\"});'")
 end
 
-def fix_pulp_httpd_conf
-  return true unless File.exist?('/etc/httpd/conf.d/pulp.conf.rpmnew')
-
-  Kafo::Helpers.execute('cp /etc/httpd/conf.d/pulp.conf.rpmnew /etc/httpd/conf.d/pulp.conf')
```
stbenjam: @ehelms @jlsherrill Do you know why we needed this and was marked an always-run?  It's causing problems now, the BZ has a full discussion of why, but pulp.conf now duplicates stuff pulp module creates so apache won't start.

jlsherrill: if we're backporting this to some magical downstream version, we might should test an upgrade with it for both the server and the capsule with this missing.

ehelms: I honestly do not recall why it was added. Speculating that when the Pulp steps in the pre upgrade were being run, they would throw an error due to missing configuration. This would put this file back, allow that migration and such to happen and then let the installer properly handle this file.

ehelms:
@stbenjam mind updating this? We'd like it for 3.2.1

bkearney:
@stbenjam ack, would like to see this get in for upgrades.

stbenjam:
Tested an upgrade yesterday and seemed OK.  Updated @ehelms, want to take another look?

PR #441, opened 2016-11-24
Remove class mappings, now correct by default

ekohl:
Based on https://github.com/theforeman/foreman-installer/commit/c7f3973e9f5dbf258130057773a00782c3937265. It also removes a stray sam-installer migration file.

This is currently untested, but here as a placeholder that it can be done. Bumping the minimum kafo version might also be required to support new Puppet 4 types used in https://github.com/theforeman/puppet-foreman_proxy/commit/af05b0791d347451f301a5995e4b150b738966b4 (and similar PRs are open for the other modules).


ehelms:
Looking good - can you get some Redmine's for these? Does Kafo figure out the mappings itself now?

ekohl:
Amended with redmine isues. And see https://github.com/theforeman/kafo/commit/4c7e7c0ce778fa4c11ef3f6c95237c40ae16fd79 for the commit that we depend on, hence the bump to at least 1.0.5.

ehelms:
Thanks @ekohl !

PR #442, opened 2016-11-28
fixes #17507 - remove workaround for ssl.conf

stbenjam:
When puppetlabs-apache is released with the fix for this, we no
longer need to worry about the RPM overwriting the ssl.conf
file.

https://github.com/puppetlabs/puppetlabs-apache/pull/1543 was accepted.

This PR mostly as a reminder to myself, it'll probably bit a little while before the change appears in a released version.


stbenjam:
No need, new puppetlabs-apache was released today.  This needs to get merged ASAP so we don't break BATS tonight.

PR #444, opened 2016-11-30
refs #17534 - fix bug with server_ssl_crl migration

stbenjam:


config/katello.migrations/161130090556-server_ssl_crl_blank.rb
```diff
@@ -1,3 +1,3 @@
 # server_ssl_crl is expected to be an absolute path or a blank string, not
-# false. Katelllo isn't using a CRL, so this needs to be a blank string.
-answers['foreman']['server_ssl_crl'] = ""
+# Katello's default value of 'false'
+answers['foreman']['server_ssl_crl'] = "" unless answers['foreman']['server_ssl_crl']
```
ehelms: If the answers file already has false, does that equate to a boolean value here or could it possibly be a string with the word false?

stbenjam: Treated as a boolean.  Could there be a case where it's quoted? I don't think so, but I can check for both if you'd prefer

PR #447, opened 2016-12-08
Fixes #16946 - add check to look for pulp ca

chris1984:
Without CA Present:

  Success!
  * Satellite is running at https://rhel6.example.com
      Initial credentials are admin / NsApaaAicBfRVeoP

  The full log is at /var/log/foreman-installer/satellite.log
[ INFO 2016-12-07 23:22:10 verbose] ssl.conf is already present, skipping
[ WARN 2016-12-07 23:22:10 verbose] Pulp CA is not present, creating.
[ INFO 2016-12-07 23:22:11 verbose] All hooks in group post finished

With CA Present:

  Success!
  * Satellite is running at https://rhel6.example.com

  The full log is at /var/log/foreman-installer/satellite.log
[ INFO 2016-12-07 23:28:31 verbose] ssl.conf is already present, skipping
[ INFO 2016-12-07 23:28:31 verbose] Pulp CA is already present, skipping
[ INFO 2016-12-07 23:28:31 verbose] All hooks in group post finished

Way to test:

install EL6 Downstream:

* "hostname foo.bar.baz", ensure "hostname -f" returns "Unknown host"
* yum install satellite (not satellite-installer)
* satellite-installer --scenario satellite
* fix hostname, re-run satellite-installer

### DIFF
--- a/hooks/post/12-check_pulp_ca.rb
+++ b/hooks/post/12-check_pulp_ca.rb
@@ -1,0 +1,8 @@
+# hook to create pulp CA on EL6 if not present
+
+if File.file?('/etc/pki/pulp/ca.crt') && File.file?('/etc/pki/pulp/ca.key')
+  logger.info 'Pulp CA is already present, skipping'
+else
+  logger.warn 'Pulp CA is not present, creating.'
+  `pulp-gen-ca-certificate`
+end


hooks/post/12-check_pulp_ca.rb
```diff
@@ -0,0 +1,8 @@
+# hook to create pulp CA on EL6 if not present
+
+if File.file?('/etc/pki/pulp/ca.crt') && File.file?('/etc/pki/pulp/ca.key') == true
```
ekohl: Isn't ```== true``` redundant?

chris1984:
@ekohl updated commit message

chris1984:
@ekohl removed the == true part

chris1984:
@beav can you answer @ekohl question

chris1984:
@beav can you answer what @stbenjam asked? I dont know enough about pulp to be able to answer this

beav:
re https://github.com/Katello/katello-installer/pull/447#issuecomment-265850270, this could be done in either a hook or in the puppet module. The puppet module may make more sense, since it's an unordered state thing.

The cert is currently created in `%post` of the pulp server rpm to help with "bare" pulp installs with no katlelo, but it can sometimes fail on el6 if the hostname isn't set correctly.

chris1984:
@stbenjam based on the last comment do you want me to close out this PR and open one under the puppet-pulp repo?

stbenjam:
Yea I think that'd be best, the change could probably go to https://github.com/Katello/puppet-pulp/blob/master/manifests/config.pp or install.pp using the [exec](https://docs.puppet.com/puppet/latest/type.html#exec) resource to run gen-ca if the file doesn't exist.

chris1984:
Closing PR based on discussion with stbenjam

PR #448, opened 2016-12-12
Fixes #17643 - Add check to verify capsule tar

chris1984:
Results of my testing:

https://paste.fedoraproject.org/505514/15784011/


hooks/pre_validations/12-check_capsule_tar.rb
```diff
@@ -0,0 +1,11 @@
+def error(message)
+  say message
+  logger.error message
+  kafo.class.exit 101
+end
+
+if param('capsule', 'certs_tar') && (certs_tar = param('capsule', 'certs_tar').value)
+  unless File.file?(certs_tar)
+    error 'The capsule tar file generated by the server is not present in the path specified, exiting.'
```
stbenjam: I'd say certs tar here

stbenjam: I'd also actually give the path that was specified, gives the user a chance to see if they fat fingered it. Something similar to 

```ruby
 error "The certs tar was not found at #{certs_tar}, exiting."
````

stbenjam:
Otherwise :+1:

chris1984:
@stbenjam thanks for the review :) Fixed the requested changes.

stbenjam:
Thanks @chris1984 !

PR #452, opened 2016-12-16
Fixes #17705 - Require port when passing proxy url

chris1984:


stbenjam:
Not a fan of this.

Our proxy configuration is weird, a port is *by definition* a part of a URI, the fact we have it as a separate option is a bug itself.  That should get fixed and the option removed.

Anyway for this, if a user gives us a url and not a port we should do the right thing and pick the default port for the protocol (80 or 443).

PR #454, opened 2016-12-16
fixes #17714 - move puppet to top level

stbenjam:
- [x] This PR
- [x] https://github.com/Katello/puppet-foreman_proxy_content/pull/108
- [x] https://github.com/Katello/puppet-certs/pull/117

Testing box (+ this script: https://gist.github.com/stbenjam/0995d0c02c72554e94d28641c793f379)

```
bye-puppet:
  box: centos7
  memory: 6144
  ansible:
    playbook: 'playbooks/katello.yml'
    group: 'server'
    variables:
        katello_version: nightly
        foreman_version: nightly
        puppet_version: 4
        foreman_installer_skip_installer: True
        puppet_repositories_version: 4
        foreman_installer_module_prs: "katello/foreman_proxy_content/108,katello/certs/117"
```

Results:

```
[root@bye-puppet ~]# foreman-installer --scenario katello --disable-system-checks
Installing             Done                                               [100%] [...........................................................................................................................................................]
  Success!
  * Katello is running at https://bye-puppet.example.com
      Initial credentials are admin / SkDQMgUZCuMr2bYZ

      foreman-proxy-certs-generate --foreman-proxy-fqdn "$FOREMAN-PROXY" --certs-tar "~/$FOREMAN-PROXY-certs.tar"

[root@bye-puppet ~]# puppet agent --test
Info: Using configured environment 'production'
Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: Caching catalog for bye-puppet.example.com
Info: Applying configuration version '1482263621'
```


ehelms:
Should https://github.com/Katello/katello-installer/pull/454/files#diff-12864807e020ec206443a1d8f1887c1dR7 be removed now?

config/foreman-proxy-content.migrations/161216113043-move_puppet_to_top_level.rb
```diff
@@ -0,0 +1,18 @@
...
+answers[:puppet][:server_foreman_url] = "https://#{answers[:foreman_proxy_content][:parent_fqdn]}"
+
+scenario[:order] = [
+  'certs',
+  'foreman_proxy',
+  'foreman_proxy::plugin::pulp',
+  'foreman_proxy_content',
+  'certs::puppet',
+  'puppet'
+]
```
ehelms: If anything else changes the order, this would over-write it. Might be safer to just append the two new items to the end of the existing order?

config/katello.migrations/161216113043-move_puppet_to_top_level.rb
```diff
@@ -0,0 +1,18 @@
...
+scenario[:order] = [
+  'certs',
+  'foreman',
+  'katello',
+  'foreman_proxy',
+  'foreman_proxy::plugin::pulp',
+  'foreman_proxy_content',
+  'certs::puppet',
+  'puppet'
+]
```
stbenjam: > If anything else changes the order

Well if anything else changes the order it should do the migration in the same way which would be run after this one.

In this case, I could just append it I guess, but not all cases of altering the order can just be a simple append.

stbenjam:
These are all ready for another look! :eyes:

ehelms:
With this change, looks we will need to advise users to regenerate certificates across their servers and capsules when upgrading?

stbenjam:
I didn't try yet, but only the path is changing - if puppet-certs doesn't redeploy them to the right location, isn't that a bug?

ehelms:
Fair point.

stbenjam:
Tested, it does the right thing in an upgrade.

config/foreman-proxy-content.migrations/161216113043-move_puppet_to_top_level.rb
```diff
@@ -0,0 +1,20 @@
+answers['foreman_proxy_content'].delete('puppet')
+
+answers['puppet'] ||= {
+  'server' => true,
```
ekohl: Should this be the value that was deleted above?

stbenjam: I suppose so, although if someone actually did set that to false a lot of things would break.

stbenjam:
This should be good to go again if @ehelms or @komidore64 you want to try the steps here:
https://github.com/Katello/katello-installer/pull/454#issue-196106412

stbenjam:
Just kidding, @ekohl and I talked on IRC and want to refactor some of it

PR #458, opened 2017-01-04
Add puppet-strings dependency

jlsherrill:
as its needed for rake pkg:generate_source

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,5 +7,6 @@ end
 
 gem 'kafo', '>= 1.0.5'
 gem 'librarian-puppet', '~> 2.2.0'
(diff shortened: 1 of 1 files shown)


Gemfile
```diff
@@ -8,4 +8,5 @@ end
 gem 'kafo', '>= 1.0.5'
 gem 'librarian-puppet', '~> 2.2.0'
 gem 'puppet', '>= 3.6.0', '< 4.0.0'
```
ekohl: While you're at it, maybe change this to ```< 5.0.0``` as well?

jlsherrill: added!

PR #460, opened 2017-01-04
3.3.0.rc1 release

jlsherrill:


Puppetfile
```diff
@@ -1,14 +1,29 @@
...
+mod 'puppetlabs/puppetdb'
+mod 'theforeman/dhcp',          :git => 'https://github.com/theforeman/puppet-dhcp'
+mod 'theforeman/dns',           :git => 'https://github.com/theforeman/puppet-dns'
+mod 'theforeman/git',           :git => 'https://github.com/theforeman/puppet-git'
+mod 'theforeman/tftp',          :git => 'https://github.com/theforeman/puppet-tftp'
+
+# Top-level modules
+mod 'theforeman/foreman',       :git => 'https://github.com/theforeman/puppet-foreman'
+mod 'theforeman/foreman_proxy', :git => 'https://github.com/theforeman/puppet-foreman_proxy'
+mod 'theforeman/puppet',        :git => 'https://github.com/theforeman/puppet-puppet'
```
ehelms: Copy/paste fail?

jlsherrill: Not sure what you mean?  

This seems to be auto-appended anytime i run   `rake pkg:generate_source` and appeared multiple times previously as i was working on doing the release.   I'm not entirely sure why it continues to be auto-appended into the file.  Is there a good reason for that?  

Should we not inlcude that here?

ehelms: Looks like you should set `FOREMAN_BRANCH` (https://github.com/Katello/katello-installer/blob/master/Rakefile#L36) to use the `1.14-stable` branch. I don't think the Puppetfile changes need to be included but we can commit them if you want.

jlsherrill: one question would be, should this be set somewhere at branch/release time? 

Rather than having to know to pass in the ENV variable every single time it gets run?

jlsherrill: updated this in the Rakefile for now

PR #461, opened 2017-01-04
Fixes #17926 - Allow discovery_smart_proxy using capsule scenario

sean797:


config/foreman-proxy-content.migrations/170104152429-add-discovery.rb
```diff
@@ -0,0 +1,6 @@
+scenario[:mapping]['foreman_proxy::plugin::discovery'] = {
```
ekohl: Mappings are no longer needed

config/foreman-proxy-content.migrations/170104152429-add-discovery.rb
```diff
@@ -0,0 +1,6 @@
...
+      :dir_name => 'foreman_proxy',
+      :params_name => 'plugin/discovery/params',
+      :manifest_name => 'plugin/discovery'
+  }
+answers['foreman_proxy::plugin::discovery'] = false if !answers['foreman_proxy::plugin::discovery']
```
ekohl: Wouldn't ```||= false``` work as well?

config/foreman-proxy-content.migrations/170104152429-add-discovery.rb
sean797: Since when? Without the mapping it doesn't work on my 3.2 install.

config/foreman-proxy-content.migrations/170104152429-add-discovery.rb
sean797: That would be better 👍

config/foreman-proxy-content.migrations/170104152429-add-discovery.rb
sean797: Ignore me, just checked the commit log, its not included in 3.2 :)

ehelms:
Should this plugin be available on the server as well?

sean797:
@ehelms yes, and it currently is based on https://github.com/Katello/katello-installer/blob/master/config/katello-answers.yaml#L56

ehelms:
That looks like the Discovery plugin itself, not the smart proxy plugin for discovery. Should the latter be available on the main server's proxy?

sean797:
Ah @ehelms thanks, I see what you mean, updated!

config/katello-answers.yaml
```diff
@@ -66,3 +66,4 @@ foreman_proxy::plugin::dhcp::infoblox: false
 foreman_proxy::plugin::dns::infoblox: false
 foreman_proxy::plugin::openscap: false
 foreman_proxy::plugin::remote_execution::ssh: false
+foreman_proxy::plugin::discovery: false
```
ehelms: Needs a migration as well.

sean797: 👍

ehelms:
@ekohl please re-review

PR #462, opened 2017-01-06
Fixes #17975 - Suggest a valid variable name for proxy-certs-generate

dLobatog:
FOREMAN-PROXY isn't a valid identifier in bash because of the dash, 
so if an user copies and pastes that command, it will never work as that
variable cannot be set. It should be something like FOREMANPROXY
or FOREMAN_PROXY so users can set that variable.

### DIFF
--- a/hooks/post/10-post_install.rb
+++ b/hooks/post/10-post_install.rb
@@ -37,7 +37,7 @@ if [0, 2].include?(@kafo.exit_code)
       say <<MSG
   * To install an additional Foreman proxy on separate machine continue by running:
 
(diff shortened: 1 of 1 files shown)


ehelms:
I never even thought of using like this, and always just assumed we output it because it was a generic identifier to make folks think about inserting the FQDN of their proxy. I assume you've tried actually setting those environment variables when calling it and everything is A-OK?

Could you add a ticket?

dLobatog:
@ehelms Yeah, using $FOREMAN_PROXY does work fine. Ticket added on the latest commit

ehelms:
Thanks @dLobatog !

PR #463, opened 2017-01-09
fixes #17639 - include grub2 tftp dir migration

stbenjam:


lzap:
LGTM, I thought we've merged the installers already...

stbenjam:
Some parts are, we use the foreman installer package for its modules, parser cache, etc but each scenario has it's own set of migrations.  That'll always be the case, unless Kafo implements some kind of hierarchy.

PR #466, opened 2017-01-09
fixes #17995, #17996 - fixes for 3.2 -> 3.3 upgrade

stbenjam:
This also fixes the prefix on some of the migrations to be consistent date format.


jlsherrill:
ACK, this fixed the issues i was seeing

PR #467, opened 2017-01-13
fixes #18068 - ensure pulp is available during puppet upgrade

Klaas-:
puppet-agent-oauth needs to be installed before stopping services

### DIFF
--- a/hooks/pre/31-upgrade-puppet.rb
+++ b/hooks/pre/31-upgrade-puppet.rb
@@ -15,10 +15,18 @@ def upgrade_puppet_package
   Kafo::Helpers.execute('yum install -y puppet-agent')
 end
 
+def start_httpd
+  Kafo::Helpers.execute('katello-service start --only httpd')
+end
+
+def remove_puppet_port_httpd
+  Kafo::Helpers.execute('sed -i "/^Listen 8140$/d" /etc/httpd/conf/ports.conf')
+end
+
 def copy_data
   success = []
   success << Kafo::Helpers.execute('cp -rfp /etc/puppet/environments/* /etc/puppetlabs/code/environments') if File.directory?('/etc/puppet/environments')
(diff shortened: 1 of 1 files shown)


stbenjam:
In the Katello upgrades we solve this kind of problem by just starting `httpd` before we run the upgrade - so pulp is available.  I wonder if that would work here instead? 

It may not work, since we remove the puppet-server package but the httpd is probably still configured for the Ruby puppet master via  Passenger.

Klaas-:
@stbenjam I did not try that, I only noticed the missing package leads to errors and you guys already have the same functionality for puppet-agent and puppetserver packages :)

Klaas-:
@beav wrote the code for the puppet-upgrade -- maybe he can tell us more :)

beav:
I'm 👍 for starting httpd before upgrading to make the packages available. That way we don't have to find the packages to install one by one via trial and error

Klaas-:
okay, so you want to close this pr and fix it yourself or do you want to merge this one and remove it once you've finished and tested if httpd starting during puppet-upgrade is working?

Klaas-:
@beav @stbenjam how are we going to move on with this bug?

stbenjam:
The best solution would be to update this PR to start httpd instead and we can help you test it.

Klaas-:
@stbenjam done, its untested as of now - I already deleted the snapshots of my update test vms

chris1984:
@Klaas-  do you have steps to test this? I want to do a test here.

Klaas-:
@chris1984 I upgrade from katello 3.1.* (puppet3) to katello 3.2.3 and after the upgrade I upgraded puppet -- just like the docs suggest.

I will have another system to test once I've finished the acceptance testing of katello 3.2.3, but I can't give an eta for that

Klaas-:
okay I changed some more things to get httpd working - I tested this revision and it seems to work

jlsherrill:
@stbenjam you okay with this?

stbenjam:
Thanks everyone!

PR #468, opened 2017-01-18
fixes #18132 - hooks in --noop delete pulp.conf

Klaas-:
This workaround is not needed anymore, katello now uses the pulp.conf
instead of 05-pulps-https.conf. therefore it doesn't need to delete or
create pulp.conf in hook because it overwrites pulp.conf within the
installer

### DIFF
--- a/hooks/pre/29-remove_package_httpd_conf.rb
+++ b/hooks/pre/29-remove_package_httpd_conf.rb
@@ -1,10 +1,13 @@
-# Some packages like pulp place configuration files in /etc/httpd/conf.d that
+# Pulp places a configuration file in /etc/httpd/conf.d that
 # contain duplicates declarations of directives that the puppetlabs-apache
-# module configured elsewhere.
+# module configured elsewhere. This is only relevant on smart proxies.
+# The main katello install overwrites the pulp.conf.
 #
 # * pulp places a pulp.conf that contains a duplicate WSGI 'pulp'
 #   daemon that 05-pulps-https.conf contains.
 #
-%w(pulp.conf).each do |file|
-  File.delete(File.join("/etc/httpd/conf.d/", file)) if File.file?(File.join("/etc/httpd/conf.d/", file))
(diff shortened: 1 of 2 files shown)


stbenjam:
This is actually still needed - but only on foreman proxies.  If you want to leave this code but change the check so it doesn't run on Katello or when in noop mode:


```ruby
if Kafo::Helpers.module_enabled?(@kafo, 'foreman_proxy_content') && !app_value(:noop)
```

Klaas-:
@stbenjam I added the noop checks and for deletion I added the check you suggested. I also tried to improve the comment a little and the reasoning for the new check.

hooks/pre/29-remove_package_httpd_conf.rb
```diff
@@ -1,10 +1,13 @@
...
-# module configured elsewhere.
+# module configured elsewhere. This is only relevant on smart proxies.
+# The main katello install overwrites the pulp.conf.
 #
 # * pulp places a pulp.conf that contains a duplicate WSGI 'pulp'
 #   daemon that 05-pulps-https.conf contains.
 #
-%w(pulp.conf).each do |file|
-  File.delete(File.join("/etc/httpd/conf.d/", file)) if File.file?(File.join("/etc/httpd/conf.d/", file))
+if Kafo::Helpers.module_enabled?(@kafo, 'foreman_proxy_content') && !app_value(:noop)
```
stbenjam: Ah sorry, mistake on my part - can you fix this to say `!Kafo::Helpers.module_enabled?(@kafo, 'katello')` instead?

foreman_proxy_content is enabled on proxies and katellos, instead check if it's not a katello.

Klaas-: done :)

stbenjam:
Thanks much, looks good to me!

stbenjam:
Thanks @Klaas- !

PR #469, opened 2017-01-23
Fixes #18196 - ask user to run foreman-tail

lzap:


ekohl:
This sounds like something the user must do and as a user I'd interpret this as something required. If it is required, then we should do it automatically for the user but otherwise I'd prefer it to be phrased as a suggestion.

dLobatog:
I think the same as @ekohl - "To monitor the progress on all related services, please start..."

lzap:
I agree, reworded to:

```
Kafo::Helpers.log_and_say :info, 'Upgrading, to monitor the progress on all related services, please do:'
Kafo::Helpers.log_and_say :info, '  foreman-tail | tee upgrade-$(date +%Y-%m-%d-%H%M).log'
```

I think it's better to use `tee` so user sees the output immediately. They can compress later and send it over to use when necessary. The idea is to have *chronological order* log of the upgrade.

And I don't think we should spawn that automatically, we will be switching to journald logging soon (I hope).

hooks/pre/30-upgrade.rb
```diff
@@ -141,7 +141,9 @@ def fail_and_exit(message)
                   'Please run --upgrade first, then --upgrade-puppet.'
   end
 
-  Kafo::Helpers.log_and_say :info, 'Upgrading...'
+  Kafo::Helpers.log_and_say :info, 'Upgrading, to monitor the progress on all related services, please do:'
+  Kafo::Helpers.log_and_say :info, '  foreman-tail | tee upgrade-$(date +%Y-%m-%d-%H%M).log'
+
```
chris1984: Code looks good. The only nitpick I have is that when testing this, it runs by really quickly can we add a sleep of 2-3 seconds so the user can see the command before the upgrade goes through? 

I did a sleep of 3 and seemed decent to copy the command for my 2nd terminal.

ekohl: Maybe it should ask for input before continuing?

lzap: I don't know, interactive will kill most tests. I'd rather keep it simple (no interactivity). Added sleep tho.

ekohl: I'm not a fan of sleep. A user who expects it will be ready but those new might not expect it and miss it. On the other hand it might be a good start we can iterate on later.

PR #471, opened 2017-01-24
Fixes #17191 - don't continue with upgrade on db migrate error

iNecas:
Before this patch, we were ignoring the results of db:migrate rake task in
pre-upgrade hook. While sometimes it helped us with issues, when the second run
of db:migrate inside the installer would finish successfully (just because of
some combination plugins or jumping though multiple versions). We should
however rather be aware of the issues and fixing them, rather than ignoring it,
with potential consequences later.

### DIFF
--- a/hooks/pre/30-upgrade.rb
+++ b/hooks/pre/30-upgrade.rb
@@ -61,9 +61,9 @@ def migrate_pulp
 end
 
 def migrate_foreman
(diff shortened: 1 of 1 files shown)


stbenjam:
@iNecas LGTM, please just make @theforeman-bot happy

stbenjam:
Thanks @iNecas! Merging, I made Mr. @theforeman-bot happy for you

iNecas:
@stbenjam++

PR #472, opened 2017-01-26
Fixes #18247 - Show initial org in proxy certs gen

chris1984:
Right now the org shown to users is Default_Organization regardless of it being that or what was set during the initial install.

Set my org at install to Katello Test:

[root@centos7 ~]# cat /etc/foreman-installer/scenarios.d/katello-answers.yaml | grep initial_organization
    initial_organization: "Katello Test"

Run of certs generate:

[root@centos7 ~]# foreman-proxy-certs-generate --foreman-proxy-fqdn test.example.com --certs-tar /root/certs.tar
Installing             Done                                               [100%] [.................................................................................................................................]
  Success!

  If you do not have the smartproxy registered to the Katello instance, then please do the following:

  1. yum -y localinstall http://centos7.example.com/pub/katello-ca-consumer-latest.noarch.rpm
  2. subscription-manager register --org "Katello Test"

  2. Copy /root/certs.tar to the system test.example.com
  3. Run the following commands on the Foreman proxy (possibly with the customized
     parameters, see foreman-installer --scenario foreman-proxy-content --help and

  foreman-installer --scenario foreman-proxy-content\
                    --foreman-proxy-content-parent-fqdn           "centos7.example.com"\
                    --foreman-proxy-register-in-foreman           "true"\
                    --foreman-proxy-foreman-base-url              "https://centos7.example.com"\
                    --foreman-proxy-trusted-hosts                 "centos7.example.com"\
                    --foreman-proxy-trusted-hosts                 "test.example.com"\
                    --foreman-proxy-oauth-consumer-key            "cLhcHAr6oA4gGAMMeokY9G7ar39caDUG"\
                    --foreman-proxy-oauth-consumer-secret         "R7e9rBjGXup7xLbctbhBF3mDseFwMgxZ"\
                    --foreman-proxy-content-pulp-oauth-secret     "vtfgkRPWvhGfoApSjqE6NwatQVF2LTis"\
                    --foreman-proxy-content-certs-tar             "/root/certs.tar"
  The full log is at /var/log/foreman-proxy-certs-generate.log

### DIFF
--- a/bin/foreman-proxy-certs-generate
+++ b/bin/foreman-proxy-certs-generate
@@ -20,20 +20,26 @@ class ForemanProxyCertsGenerate
 
   private
 
+  # rubocop: disable Metrics/MethodLength
   def generate_answers_file
+    if File.file?('./config/last_scenario.yaml')
+      scenario = YAML.load_file('./config/last_scenario.yaml').fetch(:answer_file)
+      organization = YAML.load_file("#{scenario}")['foreman']['initial_organization'] rescue "Default Organization"
+    else
+      organization = "Default Organization"
+    end
     temp_file('foreman-proxy-certs-answer',
               <<-EOF
                 certs:
                   generate: true
                   deploy: false
                   group: foreman
-                  org: Default_Organization
+                  org: #{organization.tr(' ', '_')}
                 foreman_proxy_certs: true
               EOF
              )
   end
 
-  # rubocop: disable Metrics/MethodLength
   def generate_config_file
     temp_file('foreman-proxy-certs-config',
               <<-EOF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -32,3 +32,6 @@ Metrics/PerceivedComplexity:
 
 Performance/Casecmp:
   Enabled: false
+
+Style/RescueModifier:
+  Enabled: false


chris1984:
The only thing I think to add would be to join the org name with _ so that way it matches the label Katello applies to orgs with spaces. Let me know.

bin/foreman-proxy-certs-generate
```diff
@@ -21,13 +21,14 @@ class ForemanProxyCertsGenerate
   private
 
   def generate_answers_file
+    default_org = `cat /etc/foreman-installer/scenarios.d/katello-answers.yaml | grep initial_organization | cut -d '"' -f2`
```
stbenjam: It'd be better to not shell out to do this, I think it'd be better to do [YAML.load_file](https://ruby-doc.org/stdlib-1.8.6/libdoc/yaml/rdoc/YAML.html#method-c-load_file) and get the file out of the answers hash.

chris1984:
@stbenjam Thanks for the doc link :) Found a way to get the info out of the nested hash, can we remove this cop that are causing the tests to fail?

stbenjam:
Oh rubocop :policewoman:  Just change the `gsub` to `tr` and it'll still work. I guess gsub should only get used if the first parameter is a regex.

chris1984:
@stbenjam Thanks for the review, test of new code:

[root@katello33 ~]# foreman-proxy-certs-generate --certs-tar "/root/certs.tar" --foreman-proxy-fqdn "test.redhat.com"
Installing             Done                                               [100%] [.................................................................................................................................]
  Success!

  If you do not have the smartproxy registered to the Katello instance, then please do the following:

  1. yum -y localinstall http://katello33.example.com/pub/katello-ca-consumer-latest.noarch.rpm
  2. subscription-manager register --org "Big_Org"

  2. Copy /root/certs.tar to the system test.redhat.com
  3. Run the following commands on the Foreman proxy (possibly with the customized
     parameters, see foreman-installer --scenario foreman-proxy-content --help and

  foreman-installer --scenario foreman-proxy-content\
                    --foreman-proxy-content-parent-fqdn           "katello33.example.com"\
                    --foreman-proxy-register-in-foreman           "true"\
                    --foreman-proxy-foreman-base-url              "https://katello33.example.com"\
                    --foreman-proxy-trusted-hosts                 "katello33.example.com"\
                    --foreman-proxy-trusted-hosts                 "test.redhat.com"\
                    --foreman-proxy-oauth-consumer-key            "DtNyfieVTCzbGPeuYyLAJZ6qhQWBe92v"\
                    --foreman-proxy-oauth-consumer-secret         "dhos5GTLBoBsrRowx64eRBQgHQrJVhpZ"\
                    --foreman-proxy-content-pulp-oauth-secret     "i4FrJCXhz9kUmXnpFuVNzjCUa99LmFDq"\
                    --foreman-proxy-content-certs-tar             "/root/certs.tar"
  The full log is at /var/log/foreman-proxy-certs-generate.log

bin/foreman-proxy-certs-generate
```diff
@@ -21,13 +21,14 @@ class ForemanProxyCertsGenerate
   private
 
   def generate_answers_file
+    initial_org = YAML.load_file('/etc/foreman-installer/scenarios.d/katello-answers.yaml')
```
stbenjam: A couple of things here - 

We shouldn't hard code katello-answers because of Satellite, we should read `/etc/foreman-installer/scenarios.d/last_scenario.yaml` which would be a symlink to the last sceario instead.


Tests are failing because this path won't exist on a check out of git.  So we need a lot better error handling:

- make sure the file exists and it has a value for initial_organization before reading it
- if it doesn't exist, provide a default value ("Default_Organization")

chris1984:
Will revisit this tomorrow, spent a good portion of the night hacking on it. I got one logic check in, I see where I have to move the file one to.

chris1984:
@stbenjam Are you able to take a look and see whats going on with the build tests? On my nightly box if I dont have a value for initial_org or I remove the file it does not complain and sets it to Default_Organization like you suggested. If I put the file back and fill in the value then it will return the value in the yaml file.

bin/foreman-proxy-certs-generate
```diff
@@ -1,10 +1,10 @@
-#!/usr/bin/env ruby
+#!/usr/bin/ruby
 require 'rubygems'
 require 'kafo'
 require 'tempfile'
 
 class ForemanProxyCertsGenerate
-  INSTALLER_DIR = '.'
+  INSTALLER_DIR = "/usr/share/katello-installer-base"
```
stbenjam: This can probably be left as is, the RPM packaging changes it to the production directories: https://github.com/Katello/katello-packaging/blob/master/katello-installer/katello-installer-base.spec#L100

bin/foreman-proxy-certs-generate
```diff
@@ -1,10 +1,10 @@
-#!/usr/bin/env ruby
+#!/usr/bin/ruby
```
stbenjam: I don't think this needs to be changed

bin/foreman-proxy-certs-generate
```diff
@@ -20,20 +20,33 @@ class ForemanProxyCertsGenerate
...
+    if File.file?('/etc/foreman-installer/scenarios.d/last_scenario.yaml')
+      scenario = YAML.load_file('/etc/foreman-installer/scenarios.d/last_scenario.yaml').fetch(:answer_file)
+      organization = begin
+                       YAML.load_file("#{scenario}")['foreman']['initial_organization']
+                     rescue
+                       nil
+                     end
+    else
+      organization = "Default Organization"
+      if organization.nil?
```
stbenjam: What's this check for? Organization will never be nil, you're assigning it on the line above

bin/foreman-proxy-certs-generate
```diff
@@ -20,20 +20,33 @@ class ForemanProxyCertsGenerate
...
   def generate_answers_file
-    temp_file('foreman-proxy-certs-answer',
-              <<-EOF
+    if File.file?('/etc/foreman-installer/scenarios.d/last_scenario.yaml')
+      scenario = YAML.load_file('/etc/foreman-installer/scenarios.d/last_scenario.yaml').fetch(:answer_file)
+      organization = begin
+                       YAML.load_file("#{scenario}")['foreman']['initial_organization']
+                     rescue
+                       nil
+                     end
```
stbenjam: You don't have to change this but it could be a one-liner like this:

```ruby
YAML.load_file("#{scenario}")['foreman']['initial_organization'] rescue nil

File: bin/foreman-proxy-certs-generate [hunk 7]
```diff
@@ -20,20 +20,33 @@ class ForemanProxyCertsGenerate
...
+    else
+      organization = "Default Organization"
+      if organization.nil?
+        organization = "Default Organization"
+      end
+      temp_file('foreman-proxy-certs-answer',
```
stbenjam: I think this isn't in the right place, it's part of the else block that begins on line 32.  The way the code is written, you'll only ever create the temp answers file if there's no last scenario.

bin/foreman-proxy-certs-generate
```diff
@@ -20,20 +20,33 @@ class ForemanProxyCertsGenerate
 
   private
 
+  # rubocop: disable Metrics/MethodLength
   def generate_answers_file
-    temp_file('foreman-proxy-certs-answer',
-              <<-EOF
+    if File.file?('/etc/foreman-installer/scenarios.d/last_scenario.yaml')
```
stbenjam: If you want, it would be nice if this was `./config/last_scenario.yaml`, and ditto down below on line 26.

And then open another PR to add foreman-proxy-certs-generate to this sed in the packaging:
   https://github.com/Katello/katello-packaging/blob/master/katello-installer/katello-installer-base.spec#L96

The reason the paths are like this - including the INSTALLER_DIR on line 7, is it *should* be possible to run the installer out of any directory.  It used to be that way but it's broken for a bunch of reasons - one of them is we have hard coded `/etc/foreman-installer` paths like this in some places.

We need to go back and fix it, but for now if we get it right on new stuff, that'd be great.

bin/foreman-proxy-certs-generate
chris1984: @stbenjam for this I had it this way since rubocop was complaining, can we disable the cop for this? I agree after making rubocop happy it looks awful and should be on a single line.

stbenjam: Ah ok.  Disabling that check is ok for me

bin/foreman-proxy-certs-generate
chris1984: @stbenjam ack will do on changing things for the better. Is this how the packaging pr should look like? 

sed -ri 'sX\./configX%{_sysconfdir}/foreman-installer/scenarios.dXg' config/foreman-proxy-certs-generate

stbenjam: You can just add `config/foreman-proxy-certs-generate` to the end of the sed i linked to

chris1984:
@stbenjam this one is good to go with an updated commit on changes as well as https://github.com/Katello/katello-packaging/pull/388 when you have a chance to review.

bin/foreman-proxy-certs-generate
```diff
@@ -20,20 +20,26 @@ class ForemanProxyCertsGenerate
 
   private
 
+  # rubocop: disable Metrics/MethodLength
   def generate_answers_file
+    if File.file?('./config/last_scenario.yaml')
+      scenario = YAML.load_file('./config/last_scenario.yaml').fetch(:answer_file)
+      organization = YAML.load_file("#{scenario}")['foreman']['initial_organization'] rescue nil
```
stbenjam: Actually sorry, this should probably be `rescue "Default Organization"`, if for some reason there's a problem reading the file, making it nil will cause a problem down on Line 37 later

chris1984:
@stbenjam updated commit with changes on the rescue

PR #473, opened 2017-01-30
Fixes #17192 - Show output during upgrade step run

mbacovsky:
This patch allows the installer to print output of the upgrade step being run immediately when it is available instead of collecting the output and printing it after the command finishes.

It should improve user experience during the long running upgrades by providing visual indication that upgrade is in the progress. Improvements in the output of the long running tasks will follow.

### DIFF
--- a/hooks/boot/01-helpers.rb
+++ b/hooks/boot/01-helpers.rb
@@ -33,20 +33,26 @@ class Kafo::Helpers
     def execute(commands)
       commands = commands.is_a?(Array) ? commands : [commands]
       results = []
-
       commands.each do |command|
-        output = `#{command} 2>&1`
-
-        if $?.success?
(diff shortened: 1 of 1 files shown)


ekohl:
It sounds related to https://github.com/Katello/katello-installer/pull/469, but not quite the same.

stbenjam:
It's somewhat different, we run a bunch of commands during the upgrade, and this improves things to show user output live instead of after the command exits.  We're already collecting the output.

@lzap's request is to run foreman-tail to grab log files like apache and things while the upgrade runs.

stbenjam:
Thanks @mbacovsky!

lzap:
Absolute game changer.

lzap:
To be honest, I'd love "interactive" mode when installer would ask for each step: Y/n/skip.

PR #474, opened 2017-01-30
Fixes #18307 - Remove red error on ES/gutterball

chris1984:
An example of a test run on new the new code:

https://paste.fedoraproject.org/541210/79914414/

To test I just added :always_run => true to the steps for es/gutterball so the installer will run that task.

This will only affect users who dont have Justin's PR to not run steps that are run but still for customers who have do not have that code then it will still show RED and is still causing support cases to be generated for downstream.

### DIFF
--- a/hooks/post/30-upgrade.rb
+++ b/hooks/post/30-upgrade.rb
@@ -31,13 +31,16 @@ def import_subscriptions
 end
 
 def elasticsearch_message
-  return true unless Kafo::Helpers.execute('rpm -q elasticsearch')
-
-  rpms = ['ruby193-rubygem-tire', 'tfm-rubygem-tire', 'elasticsearch', 'sigar-java', 'sigar', 'snappy-java', 'lucene4-contrib', 'lucene4']
-  message = "Elasticsearch has been removed as a dependency.  The database files can be "\
+  `rpm -q elasticsearch`
+  if $?.success?
+    rpms = ['ruby193-rubygem-tire', 'tfm-rubygem-tire', 'elasticsearch', 'sigar-java', 'sigar', 'snappy-java', 'lucene4-contrib', 'lucene4']
+    message = "Elasticsearch has been removed as a dependency.  The database files can be "\
             "removed manually with # rm -rf /var/lib/elasticsearch.  "
-  message += "Some packages are no longer needed and can be removed:  # yum erase #{rpms.join(' ')}"
-  Kafo::Helpers.log_and_say :info, message
+    message += "Some packages are no longer needed and can be removed:  # yum erase #{rpms.join(' ')}"
+    Kafo::Helpers.log_and_say :info, message
+  else
+    logger.info "Elasticsearch already removed, skipping"
+  end
 end
 
 def add_export_distributor
@@ -61,12 +64,15 @@ def set_virt_who_on_pools
 end
 
 def remove_gutterball
-  return true unless Kafo::Helpers.execute('rpm -q gutterball')
-  Kafo::Helpers.execute("yum erase -y gutterball tfm-rubygem-foreman_gutterball gutterball-certs tfm-rubygem-hammer_cli_gutterball")
-
-  ['tomcat', 'tomcat6'].each do |t|
-    gutterball_dir = "/var/lib/#{t}/webapps/gutterball"
-    Kafo::Helpers.execute("rm -rfv #{gutterball_dir}") if File.directory?(gutterball_dir)
+  `rpm -q gutterball`
+  if $?.success?
+    Kafo::Helpers.execute("yum erase -y gutterball tfm-rubygem-foreman_gutterball gutterball-certs tfm-rubygem-hammer_cli_gutterball")
+    ['tomcat', 'tomcat6'].each do |t|
+      gutterball_dir = "/var/lib/#{t}/webapps/gutterball"
+      Kafo::Helpers.execute("rm -rfv #{gutterball_dir}") if File.directory?(gutterball_dir)
+    end
+  else
+    logger.info "Gutterball already removed, skipping"
   end
 end
 


hooks/post/30-upgrade.rb
```diff
@@ -31,13 +31,15 @@ def import_subscriptions
 end
 
 def elasticsearch_message
-  return true unless Kafo::Helpers.execute('rpm -q elasticsearch')
-
+  if `rpm -q elasticsearch`.gsub!(/\A"|"\Z/, '') != nil
```
stbenjam: What's this supposed to do? It seems to always be nil.  Using elasticearch (not installed) and coreutils (installed) as an example:

```
irb(main):009:0> `rpm -q elasticsearch`
=> "package elasticsearch is not installed\n"
irb(main):010:0> `rpm -q elasticsearch`.gsub!(/\A"|"\Z/, '')
=> nil
irb(main):007:0> `rpm -q coreutils`.gsub!(/\A"|"\Z/, '')
=> nil
irb(main):008:0> `rpm -q coreutils`
=> "coreutils-8.22-18.el7.x86_64\n"
```

chris1984: @stbenjam the return true unless Kafo::Helpers.execute('rpm -q elasticsearch') was always giving back package elasticsearch is not installed, so it threw up the red error. I was trying to suppress that, I tried adding it to Kafo::Helpers.execute but it was not happy with that, that is why I put it in a if statement.

stbenjam: What's the intention of the `gsub!`?  As far as I can see it always returns nil, so nothing in this if block will ever get executed

stbenjam: The `/\A"|"\Z/` regex seems to match any line that begins or ends with a `"`, so not sure what rpm output should match that

chris1984: @stbenjam sorry about that it was supposed to be rpm -qa like this

  if `rpm -qa elasticsearch`.gsub!(/\A"|"\Z/, '') != nil

hence why i was doing the .gsub to kill off the "" on the string so it truly was nil.

With the new code:

Upgrade Step: elasticsearch_message...
[ INFO 2017-01-30 13:52:29 verbose] Upgrade Step: elasticsearch_message...
[ INFO 2017-01-30 13:52:30 verbose] Elasticsearch is already removed, skipping
Upgrade Step: remove_gutterball...
[ INFO 2017-01-30 13:52:30 verbose] Upgrade Step: remove_gutterball...
[ INFO 2017-01-30 13:52:31 verbose] Gutterball is already removed, skipping
Upgrade completed!

If that is fine let me know and I will clear up the rubocop errors and resubmit

stbenjam: I still don't see what you mean, there's no `"` in the output of any rpm commands, and regardless of whether the package is installed or not you're going to get nil - so your test isn't really valid, if elasticsearch was installed, it'd skip the steps too.

stbenjam: So after talking on IRC, I see what you're trying to do the gsub on `"` but that's not really part of the empty string - best bet would be to check rpm exit code.

something like this:

```ruby
if `rpm -q elasticsearch` && $?.success?
  # do installed stuff
else
  # skip it
end
```
```

**Comment by chris1984** [comment]
@stbenjam Its failing with these 2 rubocop errors, did you want me to find a different way to lay this out or is this something we can turn off in rubocop?

**Comment by stbenjam** [comment]
Oh I see - you can change it to be:

```ruby
`rpm -q elasticsearch`
if  $?.success?
```

Or disable the cop, either is fine for me.

chris1984:
Results of test with new code:

With Elasticsearch rpm installed:

Upgrade Step: elasticsearch_message...
[ INFO 2017-02-03 18:20:51 verbose] Upgrade Step: elasticsearch_message...
Elasticsearch has been removed as a dependency.  The database files can be removed manually with # rm -rf /var/lib/elasticsearch.  Some packages are no longer needed and can be removed:  # yum erase ruby193-rubygem-tire tfm-rubygem-tire elasticsearch sigar-java sigar snappy-java lucene4-contrib lucene4
[ INFO 2017-02-03 18:20:51 verbose] Elasticsearch has been removed as a dependency.  The database files can be removed manually with # rm -rf /var/lib/elasticsearch.  Some packages are no longer needed and can be removed:  # yum erase ruby193-rubygem-tire tfm-rubygem-tire elasticsearch sigar-java sigar snappy-java lucene4-contrib lucene4
Upgrade Step: remove_gutterball...
[ INFO 2017-02-03 18:20:51 verbose] Upgrade Step: remove_gutterball...
[ INFO 2017-02-03 18:20:51 verbose] Gutterball already removed, skipping
Upgrade completed!
[ INFO 2017-02-03 18:20:51 verbose] Upgrade completed!
[ INFO 2017-02-03 18:20:51 verbose] All hooks in group post finished

With both pkgs not installed:

Upgrade Step: elasticsearch_message...
[ INFO 2017-02-03 16:05:06 verbose] Upgrade Step: elasticsearch_message...
[ INFO 2017-02-03 16:05:06 verbose] Elasticsearch already removed, skipping
Upgrade Step: remove_gutterball...
[ INFO 2017-02-03 16:05:06 verbose] Upgrade Step: remove_gutterball...
[ INFO 2017-02-03 16:05:06 verbose] Gutterball already removed, skipping
Upgrade completed!
[ INFO 2017-02-03 16:05:06 verbose] Upgrade completed!
[ INFO 2017-02-03 16:05:06 verbose] All hooks in group post finished

hooks/post/30-upgrade.rb
```diff
@@ -31,13 +31,16 @@ def import_subscriptions
...
+  if $?.success?
+    rpms = ['ruby193-rubygem-tire', 'tfm-rubygem-tire', 'elasticsearch', 'sigar-java', 'sigar', 'snappy-java', 'lucene4-contrib', 'lucene4']
+    message = "Elasticsearch has been removed as a dependency.  The database files can be "\
             "removed manually with # rm -rf /var/lib/elasticsearch.  "
-  message += "Some packages are no longer needed and can be removed:  # yum erase #{rpms.join(' ')}"
-  Kafo::Helpers.log_and_say :info, message
+    message += "Some packages are no longer needed and can be removed:  # yum erase #{rpms.join(' ')}"
+    Kafo::Helpers.log_and_say :info, message
+  else
+    logger.info "Elasticsearch already removed, skipping"
```
ekohl: What's the difference between logger here and the ```log_and_say``` helper?

chris1984: We get a duplicate message when the installer runs, here is the code:

  else
    Kafo::Helpers.log_and_say :info, "Elasticsearch already removed, skipping"

Here is the output:

Upgrade Step: elasticsearch_message...
[ INFO 2017-02-06 09:15:43 verbose] Upgrade Step: elasticsearch_message...
Elasticsearch already removed, skipping
[ INFO 2017-02-06 09:15:43 verbose] Elasticsearch already removed, skipping

Wanting to keep it logger.info to keep the noise down in the logs/terminal.

chris1984:
@stbenjam Thanks for the review, when you get a moment can you review it please. I didn't want to remove the cop because its good to have the check, so just change the if statement around.

stbenjam:
Thanks everyone!

PR #478, opened 2017-02-06
fixes #18402 - only delete pulp.conf explicitly on a proxy

stbenjam:
!katello is not the same as a proxy with content, as the hooks
also get run by foreman-proxy-certs-generate.  This causes the pulp.conf to get deleted on a Katello when generating proxy certs.


jlsherrill:
APJ

PR #479, opened 2017-02-06
Fix 3.2

ehelms:


ekohl:
This is a *big* change. Are you sure this is what you intended?

ehelms:
Yup. The Gemfile needs to be pinned in order to generate a parser_cache properly. We used to store the modules in git because we pulled the source tarball for releases from git. However, due to the timestamps in the parser cache this is no longer feasible. So the modules and parser cache shouldn't be stored in git as they take up space and aren't necessarily representative of a given build.

stbenjam:
Do we need this anymore?

PR #480, opened 2017-02-13
fixes #18543 - explicitly delete non-katello modules from cache 

stbenjam:
We need to make sure we're deleting not just foreman things, but puppet, apache, etc module caches too.  This explicitly whitelists only the modules katello-installer is shipping, it probably needs to be a bit more dynamic

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -65,9 +65,7 @@ task :generate_parser_caches => [PARSER_CACHE_DIR] do
 
     cache = YAML.load_file(filename.to_s)
     cache[:files] = cache[:files].delete_if do |k, _|
-      k =~ /\Aforeman/ &&
-        k !~ /\Aforeman_proxy_content/ &&
-        k !~ /\Aforeman_proxy_certs/
+      !%w(certs foreman_proxy_certs foreman_proxy_content katello).include?(k)
     end
     File.open(filename.to_s, "w") do |file|
       file.write(cache.to_yaml)


ehelms:
This looks good -- could we get an issue to track it?

stbenjam:
I don't think it works quite right, and I'm going to make it a bit more dynamic so hold on on reviewing for now

stbenjam:
There we go, this is sufficient - we only need top-level modules cached, which explains why we recently saw an incompatibility - puppet was moved to top-level so we were generating a cache for it, conflicting with foreman.

Rakefile
```diff
@@ -65,9 +65,7 @@ task :generate_parser_caches => [PARSER_CACHE_DIR] do
 
     cache = YAML.load_file(filename.to_s)
     cache[:files] = cache[:files].delete_if do |k, _|
-      k =~ /\Aforeman/ &&
-        k !~ /\Aforeman_proxy_content/ &&
-        k !~ /\Aforeman_proxy_certs/
+      !%w(certs foreman_proxy_content katello puppet).include?(k)
```
ekohl: I'm surprised puppet is in here. I'd think foreman-installer ships that and thus should be wiped as well.

ehelms: Agreed, based on the initial comment I assumed puppet would be omitted here, but I guess thats the point. The current error revolves around there not being a puppet parser cache found:

```
stderr: /usr/share/gems/gems/kafo-1.0.6/lib/kafo/puppet_module.rb:69:in `parse': No Puppet module parser is installed and no cache of the file /usr/share/foreman-installer/modules/puppet/manifests/init.pp is available. Please check debug logs and install optional dependencies for the parser. (Kafo::ParserError)
        from /usr/share/gems/gems/kafo-1.0.6/lib/kafo/configuration.rb:93:in `block in modules'
        from /usr/share/gems/gems/kafo-1.0.6/lib/kafo/configuration.rb:93:in `map'
        from /usr/share/gems/gems/kafo-1.0.6/lib/kafo/configuration.rb:93:in `modules'
        from /usr/share/gems/gems/kafo-1.0.6/lib/kafo/configuration.rb:198:in `params'
        from /usr/share/gems/gems/kafo-1.0.6/lib/kafo/configuration.rb:208:in `preset_defaults_from_puppet'
        from /usr/share/gems/gems/kafo-1.0.6/lib/kafo/kafo_configure.rb:277:in `set_parameters'
        from /usr/share/gems/gems/kafo-1.0.6/lib/kafo/kafo_configure.rb:100:in `initialize'
        from /usr/share/gems/gems/clamp-1.0.0/lib/clamp/command.rb:133:in `new'
        from /usr/share/gems/gems/clamp-1.0.0/lib/clamp/command.rb:133:in `run'
        from /usr/share/gems/gems/kafo-1.0.6/lib/kafo/kafo_configure.rb:155:in `run'
        from /sbin/foreman-installer:8:in `<main>'
        to retry, use: --limit @/home/ehelms/.ansible/retry-files/nightly_server_proxy_client.retry
```

stbenjam: Yes it was the whole cause of the bug initially was puppet being in there, I have no idea why I included in the list :man_facepalming:

Rakefile
```diff
@@ -65,9 +65,7 @@ task :generate_parser_caches => [PARSER_CACHE_DIR] do
 
     cache = YAML.load_file(filename.to_s)
     cache[:files] = cache[:files].delete_if do |k, _|
-      k =~ /\Aforeman/ &&
-        k !~ /\Aforeman_proxy_content/ &&
-        k !~ /\Aforeman_proxy_certs/
+      !%w(certs foreman_proxy_certs foreman_proxy_content katello).include?(k)
```
ekohl: Where does ```foreman_proxy_certs``` come from btw?

ehelms: foreman-proxy-certs-generate -- https://github.com/Katello/katello-installer/blob/master/bin/foreman-proxy-certs-generate#L23

stbenjam:
Merged, pls2kick off build/pipeline :pray:

PR #483, opened 2017-02-23
fixes #18405 - modify upgrade to include rake tasks to fix mismatches

bbuckingham:
There exist several rake tasks in katello to resolve data integrity
issues (or mismatches) between it and the backend services
(pulp and candlepin).  This commit adds those steps to the installer
upgrade process so that the issues can be addressed automatically
versus the current behavior which may result in a failed upgrade
and then a need to run them manually.

Note: this is primarily for differences related to repository
and consumer resources.

The commit also includes an option  (--disable-resolve-mismatches)
to run the upgrade without making any changes.  It will still
run the tasks and report the changes that it would have made.

### DIFF
--- a/hooks/boot/30-upgrade.rb
+++ b/hooks/boot/30-upgrade.rb
@@ -4,3 +4,9 @@ app_option(
   "Run the steps necessary for an upgrade such as migrations, rake tasks, etc.",
   :default => false
 )
+app_option(
+  '--disable-resolve-mismatches',
+  :flag,
+  "This will disable the resolving of mismatches between the application and backend services, during upgrade.  The steps will still run in a non-commit mode to show what would have been changed.",
+  :default => false
+)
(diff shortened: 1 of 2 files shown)


hooks/boot/30-upgrade.rb
```diff
@@ -4,3 +4,9 @@
   "Run the steps necessary for an upgrade such as migrations, rake tasks, etc.",
   :default => false
 )
+app_option(
+  '--disable-resolve-mismatches',
```
stbenjam: Nitpick but for consistency sake, the Kafo convention for negative options would be `--no-resolve-mismatches`

stbenjam: Although I guess we break the convention already with `--disable-system-checks`, doesn't really matter to me

bbuckingham: @stbenjam, yeah, i based it off of the disable-system-checks, but am totally open to opinions for consistency and clarity. :)

Thanks for reviewing!!

chris1984:
Code looks good, going to run a test with an internal db for timing sake

jlsherrill:
LGTM too

bbuckingham:
@stbenjam, @chris1984, @jlsherrill, Thanks guys!

PR #485, opened 2017-03-06
fixes #18797 - Not to use relative path with certs-tar option

netman2k:
The hooks/post/10-post_install.rb file uses relative path with certs-tar option.
but I can not be used with. cause the certs-tar option requires absolute path.
The error was:

```
[root@katello01 ~]# foreman-proxy-certs-generate --foreman-proxy-fqdn "$FOREMAN_PROXY" --certs-tar "~/$FOREMAN_PROXY-certs.tar"
Parameter certs-tar invalid: ~/puppetca01.example.net-certs.tar is not one of regexes matching /^(([a-zA-Z]:[\\\/])|([\\\/][\\\/][^\\\/]+[\\\/][^\\\/]+)|([\\\/][\\\/]\?[\\\/][^\\\/]+))/ or regexes matching /^\/([^\/\0]+(\/)?)+$/
Error during configuration, exiting
```

### DIFF
--- a/hooks/post/10-post_install.rb
+++ b/hooks/post/10-post_install.rb
@@ -28,7 +28,7 @@ if [0, 2].include?(@kafo.exit_code)
       say <<MSG
   * To install an additional Foreman proxy on separate machine continue by running:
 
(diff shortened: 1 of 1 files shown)


ekohl:
Thanks!

PR #488, opened 2017-03-09
Fixes #18849: katello-certs-check does not check expiration date

Aftermath:
Added checks for the expiration date of the CA and the Certificate individually.
Had them checked separately as they can be created/developed separately in user environments.

### DIFF
--- a/bin/katello-certs-check
+++ b/bin/katello-certs-check
@@ -53,6 +53,27 @@ function success () {
     echo "[OK]"
 }
 
+function check-expiration () {
+  DATE=$(date -u +"%b %-d %R:%S %Y")
+  CERT_EXP=$(openssl x509 -noout -enddate -in $CERT_FILE | sed -e 's/notAfter=//' | awk '{$NF="";}1')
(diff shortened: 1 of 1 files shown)


bin/katello-certs-check
```diff
@@ -53,6 +53,27 @@ function success () {
     echo "[OK]"
 }
 
+function check-expiration () {
+  DATE=$(date -u +"%b %-d %R:%S %Y")
```
ekohl: Please match the indenting of other functions

Aftermath: Sorry, missed that.
I will submit that in the next commit.
Looking over some other methods of verification before I finalize.

Aftermath:
Tried to look into the feasibility of getting the expiration from concatenated CA's that may include intermediates, but this would not be simply accomplished by `openssl` commands, we would need to separate the certs out and read them individually by some bash trickery, and I don't this its completely necessary. It seems more likely that the base certificate or CA that actually signed the cert will expire prior to the intermediate or Root CA above them, making this a unlikely case.

ehelms:
Thanks @Aftermath !

PR #495, opened 2017-03-28
Fixes #19180 - Remove crane

ekohl:
The crane module is deprecated and now part of katello/pulp as pulp::crane.

I believe no migration is needed since it wasn't exposed directly to users.


ekohl:
Failure is due to rubocop.

ehelms:
@ekohl Mind doing a Redmine for this?

ehelms:
I think the Rubocop issue is fixed but a Redmine would be appreciated.

ekohl:
Done

PR #496, opened 2017-03-31
Fixes #19118 - Configure CLI on server

ehelms:
The recent change to hammer to verify SSL means we need to configure SSL and the hostname of the hammer config needs to match. The base configuration deploys at `https://localhost' and the `foreman::cli` manifest configure both the SSL certificate and sets the hostname properly. 

This would fix current bats failures -- http://ci.theforeman.org/view/Katello%20Pipeline/job/release_test_katello_nightly_el7/291/console

### DIFF
--- a/hooks/pre_migrations/01-migrate_legacy_foreman_installer_config.rb
+++ b/hooks/pre_migrations/01-migrate_legacy_foreman_installer_config.rb
@@ -13,7 +13,7 @@ if File.exists?(legacy_config_file)
     scenario = Kafo::Configuration.new(scenario_file)
 
     # copy over config and answers
-    scenario.migrate_configuration(legacy_config, :skip => [:log_dir, :log_name])
+    scenario.migrate_configuration(legacy_config, :skip => %i(log_dir log_name))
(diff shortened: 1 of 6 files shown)


stbenjam:
Looks ok to me, can we fix the rubocop failure?

hooks/pre_migrations/01-migrate_legacy_foreman_installer_config.rb
```diff
@@ -13,7 +13,7 @@
     scenario = Kafo::Configuration.new(scenario_file)
 
     # copy over config and answers
-    scenario.migrate_configuration(legacy_config, :skip => [:log_dir, :log_name])
+    scenario.migrate_configuration(legacy_config, :skip => %i(:log_dir, :log_name))
```
stbenjam: Should just be `%i(log_dir log_name)`

ehelms: I thought it looked weird. Updated

ehelms:
Rubocop friday over here. Fixed the new issue too in the Rakefile.

PR #499, opened 2017-04-14
Fixes #19276 - Stop pulp services on reset

chris1984:


hooks/pre/10-reset_feature.rb
```diff
@@ -19,8 +19,9 @@ def foreman_installed?
 end
 
 def stop_services
-  Kafo::KafoConfigure.logger.info 'Ensuring services httpd and foreman-tasks are stopped.'
-  Kafo::Helpers.execute('service httpd stop && service foreman-tasks stop')
+  Kafo::KafoConfigure.logger.info 'Stopping services'
+  Kafo::Helpers.execute('katello-service stop')
```
ehelms: The katello-service script supports a way to exclude services that could be used instead of starting postgresql in the next line.

chris1984: @ehelms I tried that and it was not working, I will file an issue for this. I went with this route for the interim.

chris1984:
@ehelms updated with katello-service

hooks/pre/10-reset_feature.rb
```diff
@@ -19,8 +19,8 @@ def foreman_installed?
 end
 
 def stop_services
-  Kafo::KafoConfigure.logger.info 'Ensuring services httpd and foreman-tasks are stopped.'
-  Kafo::Helpers.execute('service httpd stop && service foreman-tasks stop')
+  Kafo::KafoConfigure.logger.info 'Stopping services'
+  Kafo::Helpers.execute('katello-service stop --exclude postgresql'
```
ekohl: When you're doing this, does it also make sense to exclude mongodb?

chris1984: @ekohl not really because we do it all here:

https://github.com/Katello/katello-installer/blob/master/hooks/pre/10-reset_feature.rb#L46-L55

PR #500, opened 2017-04-24
fixes #18261 - Run pulp-manage-db on ostree enable

parthaa:
This hook will run pulp-manage-db on
--katello-enable-ostree=true or --katello-devel-enable-ostree=true

### DIFF
--- a/hooks/post/40-enable-katello-ostree.rb
+++ b/hooks/post/40-enable-katello-ostree.rb
@@ -1,0 +1,10 @@
+# Run pulp-manage-db if --katello-enable-ostree=true or --katello-devel-enable-ostree=true
+
+ostree_enabled = @kafo.param('katello', 'enable_ostree') ?  @kafo.param('katello', 'enable_ostree').value : false
+ostree_devel_enabled = @kafo.param('katello_devel', 'enable_ostree') ? @kafo.param('katello_devel', 'enable_ostree').value : false
+
+if ostree_enabled || ostree_devel_enabled
+  Kafo::Helpers.execute('katello-service stop --only pulp_workers,pulp_resource_manager,pulp_celerybeat')
+  Kafo::Helpers.execute('su - apache -s /bin/bash -c pulp-manage-db')
+  Kafo::Helpers.execute('katello-service start --only pulp_workers,pulp_resource_manager,pulp_celerybeat')
+end


stbenjam:
What I was thinking is *another* pulp-manage-db exec in https://github.com/Katello/puppet-pulp/blob/master/manifests/database.pp that only runs when refreshed, and set up a notify in puppet-katello when ostree is enabled.  I'd like to see what @ekohl or @ehelms thinks about that idea, not sure if it makes sense.

I'd prefer not to do things in hooks, especially since this will run in every post

ehelms:
I tend to agree with @stbenjam. Given that you can enable plugins via puppet-pulp, that module should be able to handle ensuring proper database state if a plugin is enabled. I think this might be more complicated no matter the solution with newer Pulps the workers, resource manager and celerybeat all have to be stopped for db manage to run.

parthaa:
@ehelms @stbenjam so are you saying add the exec code to also stop the services before running pulp-manage-db and starting em there after ?. Bigger issue is how is the code to know the --katello-enabled-ostree was not set to true before and suddenly is set to true ?

ekohl:
I believe https://github.com/Katello/puppet-pulp/blob/74fb5e2c9b666253ae0bc3b4465f2344b8f235d2/manifests/install.pp#L46 should be the line that knows the database should be updated. I sort of agree with @stbenjam that that should trigger another exec. Since the guard is the existence of ```/var/lib/pulp/init.flag``` (https://github.com/Katello/puppet-pulp/blob/74fb5e2c9b666253ae0bc3b4465f2344b8f235d2/manifests/database.pp#L18) my thinking was some exec that removes the file before the migration exec. It could look something like:

```puppet
exec { 'force database migration':
  command     => 'rm -f /var/lib/pulp/init.flag',
  refreshonly => true,
  before      => Exec['migrate_pulp_db'],
}
```

If other plugins need a similar mechanism then it can easily be reused.

parthaa:
@stbenjam @ekohl bigger problem here is I need to figure out how to stop pulp services. Wait for the operations to complet. Then run pulp-manage-db and then starting em back

parthaa:
@stbenjam @ekohl if you guys have any suggestions on that I 'd really appreciate.. Thanks

ekohl:
@parthaa I think we talked about this on IRC briefly but I'm wondering if it can be a pre-start exec in the systemd unit file. That way we can simply restart the service from puppet and end up with a migrated database. The only thing is that a sysadmin might not expect a startup to take long after an upgrade, but wouldn't that be an inconsistent state anyway?

stbenjam:
After discussing on IRC, one of the limitations is the pulp services must be stopped prior to a migration.  So, first thing is - I think this should be abstracted to be generic.  In Pulp, you'd add some generic define:

```puppet
define pulp::plugin(
  # we need some stuff here to customize things, like package name, etc
  $version = ...
  $package = ...
  $whatever = ...
) {

  package { "pulp-${name}-plugins":
    ensure => present,
    notify   => Exec['force-pulp-migration'],
  }

  # anything else needed for a pulp plugin

}
```


And then you can adapt @ekohl's example to do something like this:

```puppet
exec { 'force database migration':
  command     => 'rm -f /var/lib/pulp/init.flag',
  refreshonly => true,
  before      => Exec['migrate_pulp_db'],
} ~>
exec { 'stop pulp services':
  command     => 'katello-service stop --only pulp_workers,pulp_resource_manager,pulp_celerybeat',
  refreshonly => true,
}
```

That will ensure the services are stopped.  You don't actually need to start them again.  The database migration task occurs before service, so they should get started again there.

ekohl:
@stbenjam my first attempt at that was https://github.com/Katello/puppet-pulp/pull/240. It's a step in that direction, but I'm not 100% sure yet we can reach that goal.

beav:
It looks like this PR had a lot of discussion and then stopped. Is it still outstanding?

ekohl:

PR #503, opened 2017-05-11
Fixes #16299 - add check for CA flag on server crt

chris1984:


bin/katello-certs-check
```diff
@@ -74,6 +74,16 @@ function check-expiration () {
     fi
 }
 
+function check-cert-ca-flag () {
+    printf "Checking if server cert has CA:TRUE flag"
+    openssl x509 -in $CERT_FILE -noout -text | grep -o CA:TRUE
```
ekohl: Why ```-o``` and not ```-q```?

chris1984: @ekohl good point, didnt know that flag existed, will change up and resubmit. I agree quiet should be used.

jlsherrill:
Merging, thanks @chris1984 !

PR #504, opened 2017-05-12
Fixes #19563 - Remove /etc/httpd/conf.d/pulp.conf hooks

ekohl:
https://github.com/Katello/puppet-pulp/pull/239/commits/6de2ba299ab8e9f0867d5021b48c96e7bec24885 ensures pulp.conf always has a correct content. This means we no longer need hooks to do this for us.

This should not be merged before the fix is in puppet-pulp.


ekohl:
Now updated with a redmine issue and ready for review.

jlsherrill:
Thanks @ekohl

PR #505, opened 2017-05-19
fixes #19601 -Include katello_devel in parser cache

stbenjam:


Rakefile
```diff
@@ -65,7 +65,7 @@ task :generate_parser_caches => [PARSER_CACHE_DIR] do
 
     cache = YAML.load_file(filename.to_s)
     cache[:files] = cache[:files].delete_if do |k, _|
```
ekohl: Can we use ```select``` rather than ```delete_if``` with a negative? This double negative is hard to read.

stbenjam:
Good point, updated to use select

PR #506, opened 2017-05-25
Refs #19667 - Support update of Candlepin with remote DB

mbacovsky:
Call cpdb with full DB specification instead of relying on the defaults

This PR requires https://github.com/Katello/puppet-katello/pull/192

### DIFF
--- a/hooks/pre/30-upgrade.rb
+++ b/hooks/pre/30-upgrade.rb
@@ -30,7 +30,19 @@ def update_http_conf
 end
 
 def migrate_candlepin
-  Kafo::Helpers.execute("/usr/share/candlepin/cpdb --update --password #{Kafo::Helpers.read_cache_data('candlepin_db_password')}")
+  db_host = param('katello', 'candlepin_db_host').value
+  db_port = param('katello', 'candlepin_db_port').value
+  db_name = param('katello', 'candlepin_db_name').value
+  db_user = param('katello', 'candlepin_db_user').value
+  db_password = param('katello', 'candlepin_db_password').value
+  db_ssl = param('katello', 'candlepin_db_ssl').value
+  db_ssl_verify = param('katello', 'candlepin_db_ssl_verify').value
+  db_uri = "//#{db_host}" + (db_port.nil? ? '' : ":#{db_port}") + "/#{db_name}"
+  if db_ssl
+    db_uri += "?ssl=true"
+    db_uri += "&sslfactory=org.postgresql.ssl.NonValidatingFactory" unless db_ssl_verify
+  end
+  Kafo::Helpers.execute("/usr/share/candlepin/cpdb --update --database '#{db_uri}' --user #{db_user} --password #{db_password}")
 end
 
 def start_tomcat


mbacovsky:
@ekohl, @chris1984 based on comments in https://github.com/Katello/puppet-katello/pull/192 I've added a patch to support SSL for Candlepin DB connections. The new changes are in separate commit.

hooks/pre/30-upgrade.rb
```diff
@@ -30,7 +30,19 @@ def update_http_conf
...
+  db_port = param('katello', 'candlepin_db_port').value
+  db_name = param('katello', 'candlepin_db_name').value
+  db_user = param('katello', 'candlepin_db_user').value
+  db_password = param('katello', 'candlepin_db_password').value
+  db_ssl = param('katello', 'candlepin_db_ssl').value
+  db_ssl_verify = param('katello', 'candlepin_db_ssl_verify').value
+  db_uri = "//#{db_host}" + (db_port.nil? ? '' : ":#{db_port}") + "/#{db_name}"
+  if db_ssl
+    db_uri += "?ssl=true"
+    db_uri += "&sslfactory=org.postgresql.ssl.NonValidatingFactory" if db_ssl_verify
```
beav: Is the test for `db_ssl_verify` correct? it seems like it should be `if not` instead of `if`

ekohl:
Btw, would removing ```/var/lib/candlepin/cpdb_done``` work as well? There's a migrate command in the puppet-candlepin module.

hooks/pre/30-upgrade.rb
mbacovsky: omg, fixed

mbacovsky:
@ekohl I'm not sure what is the difference between calling [liquibase](https://github.com/Katello/puppet-candlepin/blob/master/manifests/database/postgresql.pp#L47-0L59) and cpdb. `cpdb update` calls liquibase internaly so should be pretty similar. `cpdb` should be preferred  way according to the [comment](https://github.com/Katello/puppet-candlepin/blob/master/manifests/database/postgresql.pp#L43-L44).

I've noticed that neither cpdb nor the puppet_candlepin's call of liquibase have support for ssl. I'll need to do some more testing.

ekohl:
@mbacovsky I think we could use cpdb by now since the linked bugzilla was closed a while ago but I haven't had time to investigate yet.

mbacovsky:
@ekohl,@beav  so after some investigation:
1. the liquibase call in puppet-candlepin could be replaced with `cpdb --create --schema-only ...` this is not idempotent so removing the `cpdb_done` won't work
1. if we add `cpdb --update ....` (which should be idempotent) to the puppet-candlepin then the candlepin migration is run with each installer invocation and could be removed from the `upgrades` script. The question is if the ordering set by upgrade script is important.
1. one concern regarding ssl support - cpdb does not support ssl. The --database parameter is for DB name only. The fact we sneak in also other params like `ssl=true` is a hack and its function depends on internal jdbc uri construction. I'll file a BZ for it. Until cpdb is fixed the only other option is to call liquibase on our own. 

I'll create new PR for puppet-candlepin with 1/ and add the hack with ssl support. Not sure about 2/ but I'd prefer to do this in separate PR and not to backport it downstream.

ekohl:
Thanks for looking into it. That sounds like a good plan.

mbacovsky:
@ekohl, @beav there is new PR adding 1/ to puppet-candlepin  https://github.com/Katello/puppet-candlepin/pull/78

iNecas:
After https://github.com/Katello/puppet-candlepin/pull/78 got in, is this PR unblocked or anything else is missing?

ekohl:
@mbacovsky can you rebase? I think the tests should pass because the failure is unrelated.

mbacovsky:
@ekohl, rebased, tests are green. Thanks!

mbacovsky:
@chris1984 , @ekohl Thanks for the reviews and approvals! I don't have the merge rights here so please someone merge it.

beav:
@mbacovsky do you mind squashing into one commit? it makes it easier if the PR has to be applied as patch, since it's one patch instead of 3

ekohl:
@beav I usually squash and merge through the GH interface.

mbacovsky:
@beav done

beav:
@mbacovsky thanks!

@ekohl ya I guess I could have just done that, I forgot about the squash button 🎃

PR #507, opened 2017-06-05
Fixes #17073: Skip puppet upgrade if already installed

ehelms:


stbenjam:
I feel like touching something like `/etc/foreman-installer/.puppet4upgrade` would be safer, a user could upgrade packages and not run the installer and they'll be stuck with p3 param values.

ekohl:
We should also consider puppet 5 migrations. Puppetlabs has published nightly repositories and promised that all code will continue to be runnable under puppet 5. I also think the paths will not change so then a 4 in the filename might be misleading as the current code will also pass for 5. See https://puppet.com/blog/full-visibility-and-control-of-your-infrastructure-new-puppet-releases and skip to Puppet 5 Platform.

ehelms:
Updated per your suggestion @stbenjam

hooks/pre/31-upgrade-puppet.rb
```diff
@@ -5,6 +5,13 @@ def puppet4_available?
   !success.include?(false)
 end
 
+def puppet4_installed?
```
ekohl: Is this still used anywhere?

hooks/pre/31-upgrade-puppet.rb
```diff
@@ -74,5 +82,6 @@ def fail_and_exit(message)
   upgrade_step :remove_puppet_port_httpd
   upgrade_step :start_httpd
 
+  File.open('/etc/foreman-installer/.puppet_4_upgrade', 'w') { |file| file.write("completed on #{Time.now}") }
```
stbenjam: Based on @ekohl's other comment I think we can change this (and the text below) to just say puppet upgrade to future-proof it so users can go directly from Puppet 3 to 5.

ehelms:
Thanks for the comments, updated

hooks/pre/31-upgrade-puppet.rb
```diff
@@ -47,6 +47,7 @@ def fail_and_exit(message)
 end
 
 if app_value(:upgrade_puppet)
+  fail_and_exit 'Puppet already installed and upgraded. Skipping.' if File.exist?('/etc/foreman-installer/.puppet_4_upgrade')
```
stbenjam: This needs updating to the new file name

ekohl: I'd prefer if the file location would be a constant somewhere. That removes the risk of mismatches.

ehelms:
@ekohl @stbenjam Updated to make the file a constant

PR #513, opened 2017-06-16
Fixes #23002 - Ensure service-wait is always executable

ehelms:
(cherry picked from commit f50d51a9a514ba6292be5770f6c77e682de82c27)


ekohl:
Looks like it missed my comment here about using native file utils to chmod rather than shelling out.

PR #514, opened 2017-06-19
Fixes #20054 - Check server cert's encoding is PEM

phayz:
This pull request adds a new function, `check-cert-encoding`. It first tests if the certificate is PEM encoded, in which case it is considered a pass, output as `[OK]`. Otherwise it's considered a failure, output as `[Fail]`, with accompanying messages.


bin/katello-certs-check
```diff
@@ -112,6 +112,24 @@ function check-ca-bundle () {
     fi
 }
 
+function check-cert-encoding () {
+    printf "Checking server certificate's encoding: "
+    openssl x509 -inform per -in $CERT_FILE -noout -text &> /dev/null
```
evgeni: shouldn't this be `-inform pem` instead of `per`? That's at least what https://wiki.openssl.org/index.php/Manual:X509(1) suggests.

bin/katello-certs-check
```diff
@@ -112,6 +112,24 @@ function check-ca-bundle () {
...
+    if [ $? == "0" ]; then
+        success
+    else
+        openssl x509 -inform der -in $CERT_FILE -noout -text &> /dev/null
+        if [ $? == "0" ]; then
+            error 8 "The server certificate is in DER encoding, which is not
+            compatible. Convert the certificate to PEM encoding,
+            then test it again."
```
sean797: Not required, but maybe also add how to do that? 

```
openssl x509 -inform der -in certificate.cer -out certificate.pem

phayz: Thanks, @sean797. I originally had the `openssl` command included in the output, but it was being run, rather than simply being displayed. At least that was my theory at the time. I will try to add this, as it's much more user-friendly to provide the exact command, instead of stating "Fix it".

ekohl:
@phayz should this be merged as it is now or do you intend to change the message?

phayz:
WIP - I consider the source branch for this PR still a work in progress as I want to enhance it as follows:
1. Provide the command required to convert the server's certificate to PEM format.
2. Have the output of the `check-cert-encoding` function "wrapped", instead of being forcefully wrapped by the terminal emulator in use.

ekohl:
Looks like you need to rebase. Please do give a shout when this is ready for review again.

phayz:
**NOTE**: This branch is still a work in progress. I will fix the commit messages to comply with the Foreman bot's requirements.

The remaining enhancement planned for this branch is to output the required `openssl` command to convert the certificate from DER encoding to PEM encoding.

phayz:
I have opted to close this PR. I will raise another when my changes are ready.

PR #515, opened 2017-06-20
fixes #20056 - deletes queue even when its not created with hostname

upadhyeammit:
This handles situation where system's hostname has been changed and :event queue was created with different hostname.


upadhyeammit:
Creating new pull request for this one to fix errors.

PR #516, opened 2017-06-21
Fixes #20056 - deletes queue when its not created with hostname

upadhyeammit:
Deletes queue when :event queue was not created by system's hostname.


ekohl:
At first glance it looks correct, but I'm not very familiar with this part. @ehelms or @chris1984?

ekohl:
Thanks @upadhyeammit for the fix and @chris1984 for the review!

PR #517, opened 2017-06-30
Fixes #18301 - Reset Puppet attributes during Puppet 4 upgrade

ehelms:
When we moved puppet to the top level we broke the upgrade as these values were inherently reset by how we masked puppet inside puppet-capsule previously.

Thanks to https://github.com/theforeman/foreman-installer/pull/219 for pointing me in the right direction


hooks/pre/31-upgrade-puppet.rb
```diff
@@ -66,6 +66,29 @@ def fail_and_exit(message)
...
+    Kafo::Helpers.reset_value(param('puppet', 'server_common_modules_path'))
+    Kafo::Helpers.reset_value(param('puppet', 'server_default_manifest_path'))
+    Kafo::Helpers.reset_value(param('puppet', 'server_dir'))
+    Kafo::Helpers.reset_value(param('puppet', 'server_envs_dir'))
+    Kafo::Helpers.reset_value(param('puppet', 'server_external_nodes'))
+    Kafo::Helpers.reset_value(param('puppet', 'server_jruby_gem_home'))
+    Kafo::Helpers.reset_value(param('puppet', 'server_package'))
+    Kafo::Helpers.reset_value(param('puppet', 'server_puppetserver_dir'))
+    Kafo::Helpers.reset_value(param('puppet', 'server_puppetserver_vardir'))
+    Kafo::Helpers.reset_value(param('puppet', 'server_puppetserver_version'))
```
ekohl: Looks like I missed this one in yhttps://github.com/theforeman/foreman-installer/pull/219. I do use a hash there and I wonder if that's more readable than this long list.

ehelms: I can swap to that style if you have a strong opinion. I can sorta see it both ways.

ekohl: No strong opinion. Since you have to split up those lines it doesn't become much more readable

lpramuk:
Resetting attributes during P4 upgrade works fine.
But then I get the same error during regular P4 installer run (already upgraded)

I'm saying we need also a fix for regular runs of already upgraded P4 installer

ekohl:
https://github.com/Katello/katello-installer/commit/91289aafe23677216e1ce05ef87411ec580fdbb6 should fix that.

PR #519, opened 2017-07-12
Refs #20239 - update backup path in script

chris1984:
@ehelms Can I haz review plz


evgeni:
`rm -f` should be sufficient, but LGTM!

chris1984:
@evgeni removed the -r flag so now just -f

evgeni:
:+1:

PR #522, opened 2017-07-26
Fixes #15905 - Add forward/reverse dns check

chris1984:
Adding to pre hooks after talks with @stbenjam because in support when setting up a customer reproducer we will need to use fake dns for an install to complete. Having it in the check will prevent us from using the --disable-system-checks override flag.

### DIFF
--- a/checks/dns.rb
+++ b/checks/dns.rb
@@ -1,0 +1,30 @@
+#!/usr/bin/env ruby
+# Check to verify forward / reverse dns matches hostname
+
+def error_exit(message, code)
+  $stderr.puts message
+  exit code
+end
+
+@hostname = `hostname -f`.chomp
+@ip = `getent ahosts #{@hostname} | awk '/STREAM/ { print $1 }'`.chomp
+@reverse = `getent hosts #{@ip} | awk '{ print $2 }'`.chomp
+
+FORWARD = "Unable to resolve forward DNS for #{@hostname}"
+
+POINTS = "Forward DNS points to #{@ip} which is not configured on this server"
+
+REVERSE = "Reverse DNS #{@reverse} does not match hostname #{@hostname}"
+
+if @ip.empty?
+  error_exit(FORWARD, 2)
+end
+
+`ip -o a | awk '/ #{@ip}\\// { print $4 }' | cut -d/ -f1`.chomp
+unless $?.success?
+  error_exit(POINTS, 2)
+end
+
+unless @hostname == @reverse
+  error_exit(REVERSE, 2)
+end


chris1984:
Test failure is unrelated.

hooks/pre/13-dns_check.rb
```diff
@@ -0,0 +1,21 @@
+# Check to verify forward/reverse dns matches hostname
+
+if app_value(:disable_system_checks)
+  logger.warn 'Skipping system checks.'
+else
+  hostname = `hostname -f`
+  ip = `host #{hostname} | awk '{print $4}'`
```
ekohl: I've been told using ```host``` is considered bad practice because it's unpredictable. Please use ```getent hosts #{hostname} | awk '/STREAM { print $1 }'``` to the get IP address using the exact same logic as the OS. Note this can be IPv6 but I think that's good here since that is what the OS would try to connect to anyway. If you want to explicitly test all IPs (which is even better) you can use ```getent ahosts #{hostname} | awk '/STREAM/ { print $1 }'``` to get both IPv4 and IPv6.

hooks/pre/13-dns_check.rb
```diff
@@ -0,0 +1,21 @@
...
+  fqdn = `host #{hostname} | awk '{print $1}'`
```
ekohl: Similarly you can avoid ```host``` by using the system logic to resolve: ```getent hosts #{ip} | awk '{ print $2 }'```

hooks/pre/13-dns_check.rb
```diff
@@ -0,0 +1,21 @@
+# Check to verify forward/reverse dns matches hostname
+
+if app_value(:disable_system_checks)
+  logger.warn 'Skipping system checks.'
+else
+  hostname = `hostname -f`
```
ekohl: Generally using ```$()``` is better since you can also nest it.

hooks/pre/13-dns_check.rb
GAZDOWN: Hello, wouldn't it be better to use Sockets and rubydns, rather than calling directly system commands which may change output at any particular time?

ekohl: Given getent is part of glibc I'm not that worried about it suddenly changing but pure ruby code can also work just as well.

chris1984: @GAZDOWN were not using sockets, this method works fine.

chris1984:
@ekohl can I get another review when you get a chance, please?

hooks/pre/13-dns_check.rb
```diff
@@ -0,0 +1,21 @@
+# Check to verify forward/reverse dns matches hostname
+
+if app_value(:disable_system_checks)
+  logger.warn 'Skipping system checks.'
+else
+  hostname = `$(hostname -f)`
```
ekohl: I think this is redundant: backticks execute shell and ```$()``` is a bash subshell so you can remove the subshell.

hooks/pre/13-dns_check.rb
```diff
@@ -0,0 +1,21 @@
...
+  ip = getent hosts `#{hostname} | awk '/STREAM { print $1 }'`
```
ekohl: The backtick is in the incorrect place.

hooks/pre/13-dns_check.rb
```diff
@@ -0,0 +1,21 @@
...
+  fqdn = `getent hosts #{ip} | awk '{ print $2 }'`
+  if hostname == fqdn
+    puts "Forward DNS matches, checking IP"
```
ekohl: A ```logger.debug``` might be better (and below too).

hooks/pre/13-dns_check.rb
```diff
@@ -0,0 +1,21 @@
+# Check to verify forward/reverse dns matches hostname
+
+if app_value(:disable_system_checks)
+  logger.warn 'Skipping system checks.'
+else
+  hostname = `$(hostname -f)`
+  ip = getent hosts `#{hostname} | awk '/STREAM { print $1 }'`
+  fqdn = `getent hosts #{ip} | awk '{ print $2 }'`
+  if hostname == fqdn
```
ekohl: This is more of a general style and maybe my python side shining through. From the [Zen of Python](http://legacy.python.org/dev/peps/pep-0020/): ```Flat is better than nested```. More concrete I think the general flow is easier to understand when it's flat. In the following code I also added another check and rephrased it a bit. This is untested.

```ruby
ip = `getent hosts #{hostname} | awk '/STREAM { print $1 }'`
if ip.empty?
  logger.error "Unable to resolve forward DNS for '#{hostname}'"
  kafo.class.exit(1)
end

`ip a | grep -q #{ip}`
unless $?.success?
  logger.error "Forward DNS points to '#{ip}' which is not configured on this server"
  kafo.class.exit(1)
end

reverse = `getent hosts #{ip} | awk '{ print $2 }'`
unless hostname == reverse
  logger.error "Reverse DNS '#{reverse}' does not match hostname '#{hostname}'"
  kafo.class.exit(1)
end
```

chris1984:
Verified it works correctly:

[ERROR 2017-10-20 16:18:53 verbose] Reverse DNS 'dhcp182-167.gsslab.rdu2.redhat.com' does not match hostname 'sat.katellolabs.lan'
[ INFO 2017-10-20 16:18:53 verbose] Installer finished in 1.931920651 seconds

[root@sat ~]# hostname -f
sat.katellolabs.lan
[root@sat ~]# cat /etc/hosts
127.0.0.1   localhost localhost.localdomain localhost4 localhost4.localdomain4
::1         localhost localhost.localdomain localhost6 localhost6.localdomain6
#10.10.182.167 sat.katellolabs.lan

ehelms:
I think we should move this check to the init hooks. They are the second hook after boot, and now there are actions that can happen in init (such as puppet upgrade) that we would not want to execute if this system check failed.

ehelms:
Generally code looks good, I'll leave final review for @ekohl

chris1984:
I remember talking to Stephen about this and at first we agreed to have it in the checks section but then it would be hard for support to build a customer reproducer because we have to have the fqdn the same as the customers and it would have that gsslab extension since all the boxes are built on that network unless built locally via libvirt. 

I can move it if you want, but that was my reasoning behind putting it in a hook instead.

ekohl:
I'm not sure I follow. There's an option in the installer to skip system checks (IIRC ```-s``` and a long form). Would that be different?

chris1984:
@ekohl putting the dns in checks i get this https://paste.fedoraproject.org/paste/rcixsYXWPpT-hSqDU6uyKg can we just keep it in pre so we have an option to skip the check to build reproducers?

ekohl:
If you do it in checks then there's no need for the ```app_value``` because kafo has a built in way to skip the checks. See https://github.com/Katello/katello-installer/blob/master/checks/hostname.rb. You will need the shebang to tell it's ruby because checks are just executed standalone. Using ```logger``` and ```kafo``` can't be used either so I'd recommend to use the ```error_exit``` function like ```hostname.rb```.

A bit more background: kafo has checks built in (https://github.com/theforeman/kafo#system-checks) but katello-installer started the pattern ```--disable-system-checks```. This is something foreman-installer doesn't have. It started because of some shortcomings of checks where you can't read the scenario or answers but a forward/reverse dns check doesn't need that since it's a standalone system check.

When the katello-installer merge into foreman-installer will happen we'll need to figure out how to deal with this and modify kafo to handle the more advanced cases.

chris1984:
Works as a check now:

~~~
[root@sat checks]# foreman-installer --scenario satellite
Reverse DNS dhcp182-167.gsslab.rdu2.redhat.com does not match hostname sat.katellolabs.lan
Your system does not meet configuration criteria
[root@sat checks]# ll
total 20
-rwxr-xr-x. 1 root root  712 Oct 19 10:15 disk_size.rb
-rwxr-xr-x. 1 root root  749 Oct 24 13:51 dns.rb
-rwxr-xr-x. 1 root root  267 Oct 19 10:15 env_variables.rb
-rwxr-xr-x. 1 root root 1006 Oct 19 10:15 hostname.rb
-rwxr-xr-x. 1 root root  214 Oct 19 10:15 lang.rb
~~~

chris1984:
@ekohl can I get another review please?

checks/dns.rb
```diff
@@ -0,0 +1,30 @@
...
+
+POINTS = "Forward DNS points to #{@ip} which is not configured on this server"
+
+REVERSE = "Reverse DNS #{@reverse} does not match hostname #{@hostname}"
+
+if @ip.empty?
+  error_exit(FORWARD, 2)
+end
+
+`ip a | grep #{@ip} | awk '{print $2 }' | rev | cut -c 4- | rev`.chomp
```
ekohl: Is the cut intended to strip off the netmask? In that case I'd recommend to let cut split on the slash. You can also filter in awk:
```bash
ip -o a | awk '/ #{@ip}\// { print $4 }' | cut -d/ -f1
```

Note I used ```ip -o``` which is the programmatic output. I also modified the search slightly so that it has boundries. This prevents matching ```10.0.0.10``` if you're looking for ```10.0.0.1```.

ekohl:
It'd be great if you submitted the same check to foreman-installer.

PR #524, opened 2017-08-07
Fixes #20245 - run orphan product upgrade script

jlsherrill:


ekohl:
I wonder if there could be a generic rake task to do the upgrades rather than list every upgrade in the installer.

jlsherrill:
it would be possible, the installer currently keeps knowledge about which tasks have been run to not run them more than once (unless :run_always is set), so that knowledge would have to moved there as well (and likely those cache files migrated)

ekohl:
The reason I ask is that if katello itself knows it, it can be reused in containers. If it's generic, it can be implemented in Foreman core and other plugins could use the same.

PR #525, opened 2017-08-08
Refs #20239 - Remove old jrnl dir after upgrade

chris1984:


pmoravec:
+1 , that directory is a relict from the qpid-cpp-server-linearstore 0.30 -> 0.34 upgrade / directory structure layout migration. Currently the directory (after the current migration) contains "just" empty subdirectories (one subdirectory for each durable queue from the time of migration/qpid upgrade). So it has no big deal but makes the directory structure confusing and this directory ridiculous.

chris1984:
Build failures are unrelated

chris1984:
Opening a new PR with updated changes from downstream regarding upgrades will add this into that one.

PR #527, opened 2017-08-10
Refs #17214 - update copy path on certs generate

chris1984:


chris1984:
@evgeni updated the path, if you want to review.

chris1984:
@evgeni how does it look now

hooks/post/10-post_install.rb
```diff
@@ -52,7 +52,7 @@ def new_install?
   Once this is completed run the steps below to start the smartproxy installation:
 
   1. Ensure that the foreman-installer-katello package is installed on the system.
-  2. Copy <%= color("#{certs_tar}", :info) %> to the system <%= color("#{foreman_proxy_fqdn}", :info) %>
+  2. Copy the following file <%= color("#{certs_tar}", :info) %> to the system <%= color("#{foreman_proxy_fqdn}", :info) %> at the following location <%= color("#{File.join('/root', File.basename(certs_tar))}", :info) %>
```
ekohl: I'm not sure I like the ```the following file``` addition. Other than that you can add a suggestion: ```scp SOURCE PROXY:DEST```. While it may not work on every system, it does clearly show the intent.

ehelms: I can see the general suggestion being useful in addition?

stbenjam: I think keeping the original text and adding a copy-and-pasteable scp command suggestion would be better

```erb
scp <%= certs_tar %> root@<%= foreman_proxy_fqdn %>:/root/<%= certs_tar %>
```

ekohl: @stbenjam that was my thinking yes

stbenjam: Ah right, totally overlooked that part of your comment I guess

chris1984:
@ekohl care to look over the new addition?

ehelms:
LGTM

ehelms:
@ekohl or @stbenjam care to vet?

hooks/post/10-post_install.rb
ekohl: Would ```s/following //``` and ```s/the following location //``` make the sentence easier to read?

hooks/post/10-post_install.rb
```diff
@@ -52,7 +52,8 @@ def new_install?
...
+  scp <%= certs_tar %> root@<%= foreman_proxy_fqdn %>:/root/<%= certs_tar %>
```
ekohl: Should this do a ```basename``` like the line above it?

chris1984:
@ekohl how does this look? I am still a little unclear on the substitute of the wording, it reads fine to me.

PR #529, opened 2017-08-11
Fixes #18301 - Reset values don't persist in answers

ehelms:


ehelms:
Note this had to move before any validations or run, or it has the potential to wipe the value out into a state of brokenness. This uses `unset_value` from Kafo itself which sets `value_set` to false to force Kafo to use the default value and not any value coming from the answers file.

ehelms:
To test, run Puppet 4 upgrade as is today, and then:

```
# foreman-installer -h | grep -e foreman-proxy-puppet-ssl -e{vardir,ssldir}
```

You'll note that it does not update the answers file to the Puppet 4. Apply this patch, re-run and it should update to the right values.

ekohl:
I swear I actually tried this but it must be that it now runs before validations.

ehelms:
@ekohl the original change? or you tried this for Foreman?

@beav and/or @chris1984 would you mind reviewing/testing this?

ekohl:
@ehelms I tried implementing the same change in foreman-installer with reset_value

ehelms:
@ekohl its verrry subtle, since the current way does upgrade Puppet within the puppet modules, just not the answers file

chris1984:
testing now

chris1984:
@ehelms still having issues, the installer still tries to install puppet 3

ehelms:
@chris1984 This is ready to test again. I had to make some changes that I attempt to explain in the commit message. @ekohl You may also find the hoops I had to jump through interesting in this update.

chris1984:
Test Results Pass:

  Success!
  * Katello is running at https://centos7.example.com

      foreman-proxy-certs-generate --foreman-proxy-fqdn "$FOREMAN_PROXY" --certs-tar "/root/$FOREMAN_PROXY-certs.tar"

[ INFO 2017-08-24 13:57:20 verbose] pulp.conf is already present, skipping

Success!
katello-service restart finished successfully!
Puppet upgrade completed!
[ INFO 2017-08-24 14:02:32 verbose] Puppet upgrade completed!
[ INFO 2017-08-24 14:02:32 verbose] All hooks in group post finished
[ INFO 2017-08-24 14:02:32 verbose] Installer finished in 528.848617844 seconds

    --puppet-ssldir               Override where SSL certificates are kept. (current: "/etc/puppetlabs/puppet/ssl")
    --puppet-vardir               Override the puppet var directory. (current: "/opt/puppetlabs/puppet/cache")

ekohl:
I'll have a look at integration the puppet upgrade in foreman-installer again based on what you found.

PR #530, opened 2017-08-14
Fixes #20587 - change importing subs to general task

johnpmitsch:


johnpmitsch:
requires https://github.com/Katello/katello/pull/6885

ekohl:
Will this break users upgrading from > 2.2 but < 3.3?

johnpmitsch:
@ekohl It shouldn't, the same functionality will run, we just had the same task under different names [here](https://git.io/v77aE) and [here](https://github.com/Katello/katello/blob/master/lib/katello/tasks/upgrades/3.3/import_subscriptions.rake) which will be moved to single task `katello:import_subscriptions` in [this PR](https://github.com/Katello/katello/pull/6885)

ekohl:
@johnpmitsch I was under the impression that these upgrade steps were only run once. That would imply that if a user ran the 2.2 step with 3.2 and then upgrades to 3.5 the old virt_who step isn't run and you end up with an inconsistent state. I base this on the ```!step_ran?(step)``` part in https://github.com/Katello/katello-installer/blob/a3d3a9c4d99bf50b6bbdf2b0f68ecda57acda0e7/hooks/post/30-upgrade.rb#L121

hooks/post/30-upgrade.rb
```diff
@@ -169,7 +165,6 @@ def fail_and_exit(message)
       upgrade_step :update_subscription_facet_backend_data, :long_running => true
       upgrade_step :remove_gutterball
       upgrade_step :remove_event_queue
-      upgrade_step :set_virt_who_on_pools, :long_running => true
```
ehelms: Why delete this upgrade step?

johnpmitsch:
@ekohl @ehelms updated

hooks/post/30-upgrade.rb
```diff
@@ -51,7 +51,7 @@ def import_puppet_modules
 end
 
 def import_subscriptions
```
ekohl: I think you can remove this one since it's duplicate for older users. Users who already ran it won't have problems anyway.

hooks/post/30-upgrade.rb
```diff
@@ -84,13 +84,17 @@ def update_subscription_facet_backend_data
 end
 
 def set_virt_who_on_pools
```
ekohl: Same here

hooks/post/30-upgrade.rb
ehelms: I don't think we can safely remove these. I think you are expecting users to upgrade to every successive version on their way along the way (e.g. 3.2 -> 3.3 ->3.4) however, we have no requirement on that and have users that will go 3.2 -> 3.4.

johnpmitsch: @ekohl @ehelms is there a consensus? wondering what to do about this PR

ekohl: @ehelms since the 3.4 upgrade will execute all these steps, won't that essentially be the same as including 3.3?

johnpmitsch: @ehelms ping re ^

ehelms: Yes - assuming they are the same task.

ekohl:
I think this is correct. @ehelms could you have a look again?

johnpmitsch:
@ehelms ping

PR #532, opened 2017-08-14
Fixes #20593 - Prevent puppet upgrade on fresh install

ehelms:
The idea behind this change:

 1) Only create the upgrade complete file when the post puppet upgrade step finishes
 2) Add a check for puppet 4 installed, no --upgrade-puppet parameter passed, and create the puppet upgrade file

This last point is to try to prevent the case of installing with Pupept 4 and then being able to run --upgrade-puppet. Please check my logic adjustments carefully.

### DIFF
--- a/hooks/boot/01-helpers.rb
+++ b/hooks/boot/01-helpers.rb
@@ -10,7 +10,7 @@ class Kafo::Helpers
       mod.enabled?
     end
 
-    def log_and_say(level, message)
+    def log_and_say(level, message, do_say = true)
       style = case level
               when :error
                 'bad'
@@ -22,7 +22,7 @@ class Kafo::Helpers
 
       # \ and ' characters could cause trouble in ERB, make sure to escape them
       escaped_message = message.gsub('\\', '\\\\\\').gsub("'", %q{\\\'})
-      say "<%= color('#{escaped_message}', :#{style}) %>"
+      say "<%= color('#{escaped_message}', :#{style}) %>" if do_say
       Kafo::KafoConfigure.logger.send(level, message)
     end
 
@@ -30,26 +30,26 @@ class Kafo::Helpers
       YAML.load_file("#{puppet_dir}/foreman_cache_data/#{param}")
     end
 
-    def execute(commands)
+    def execute(commands, do_say = true)
       commands = commands.is_a?(Array) ? commands : [commands]
       results = []
       commands.each do |command|
-        results << execute_command(command)
+        results << execute_command(command, do_say)
       end
       !results.include? false
     end
 
-    def execute_command(command)
+    def execute_command(command, do_say)
       process = IO.popen("#{command} 2>&1") do |io|
         while line = io.gets
           line.chomp!
-          log_and_say(:debug, line)
+          log_and_say(:debug, line, do_say)
         end
         io.close
         if $?.success?
-          log_and_say(:debug, "#{command} finished successfully!")
+          log_and_say(:debug, "#{command} finished successfully!", do_say)
         else
-          log_and_say(:error, "#{command} failed! Check the output for error!")
+          log_and_say(:error, "#{command} failed! Check the output for error!", do_say)
         end
         $?.success?
       end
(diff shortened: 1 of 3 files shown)


chris1984:
when the installer runs without and special flags besides -v and turning off the system checks the rpm -q commands are being executed:

  Success!
  * Katello is running at https://centos7.example.com

      foreman-proxy-certs-generate --foreman-proxy-fqdn "$FOREMAN_PROXY" --certs-tar "/root/$FOREMAN_PROXY-certs.tar"

[ INFO 2017-08-16 14:58:33 verbose] pulp.conf is already present, skipping
puppet-agent-1.10.6-1.el7.x86_64
rpm -q puppet-agent finished successfully!
puppetserver-2.7.2-1.el7.noarch
rpm -q puppetserver finished successfully!
[ INFO 2017-08-16 14:58:33 verbose] All hooks in group post finished
[ INFO 2017-08-16 14:58:33 verbose] Installer finished in 35.968015535 seconds

ehelms:
@chris1984 I think that output should be expected. Depends on how we want to think about it. That is debug output, which -v enables since the default log level is debug. So we are seeing debug output. If running just the installer, without -v I don't think you'd see this output. That being said, I think we can be smarter about this. I moved the installed check to be the last check so that if either of the previous conditions are not met, it should stop there due to the `&&` and only run the `rpm -q` in a very specific set of instances -- https://github.com/Katello/katello-installer/pull/532/files#diff-b577779a7038271ea54f6f431614d9c7R50

ehelms:
So I updated this to modify the `log_and_say` function to optionally only log. That should help prevent the nasty output at the end of this run.

hooks/boot/01-helpers.rb
```diff
@@ -10,7 +10,7 @@ def module_enabled?(kafo, name)
       mod.enabled?
     end
 
-    def log_and_say(level, message)
+    def log_and_say(level, message, log_only = false)
```
ekohl: Would it be better to have a ```log(level, message)``` and ```say(level, message)``` where ```log_and_say(level, message)``` composes them? They don't appear to share variables.

ehelms: I thought about that a bit, it seemed logical but that it adds furthering branching if/else elsewhere and I couldn't decide if the cost was worth it. So thats me saying I could go either way if @ekohl and @chris1984 have a preference.

ekohl: I noticed that later on it'd be a bit awkward. I'd at least prefer ```say=true``` rather than ```log_only=false``` for the parameter.

ehelms: @ekohl so, the problem there is the function is named `say` which means I can't name the variable that as well.

ekohl: ```do_say```? Still not pretty but I think it's better than ```log_only```.

ehelms:
Updated to use `do_say` parameter.

chris1984:
P4 works fine and does not show the check anymore:

[root@centos7 ~]# foreman-installer --scenario katello --disable-system-checks
Installing             Done                                               [100%] [................................................................................................................................]
  Success!
  * Katello is running at https://centos7.example.com
      Initial credentials are admin / iiPF9vWzJApmJw9a

      foreman-proxy-certs-generate --foreman-proxy-fqdn "$FOREMAN_PROXY" --certs-tar "/root/$FOREMAN_PROXY-certs.tar"

[root@centos7 ~]# foreman-installer --scenario katello --disable-system-checks --upgrade-puppet
Puppet already installed and upgraded. Skipping.

P3 Tests:
Installer run didnt show rpm check:

Installing             Done                                               [100%] [................................................................................................................................]
  Success!
  * Katello is running at https://centos7.example.com
      Initial credentials are admin / GvYnVkDWY3spqbKW

      foreman-proxy-certs-generate --foreman-proxy-fqdn "$FOREMAN_PROXY" --certs-tar "/root/$FOREMAN_PROXY-certs.tar"

[root@centos7 ~]# foreman-installer --upgrade-puppet --disable-system-checks
Puppet already installed and upgraded. Skipping.

PR #535, opened 2017-08-30
Refs #14956 - Add katello proxy reset to installer

chris1984:


hooks/boot/13-katello_proxy_reset.rb
```diff
@@ -0,0 +1,7 @@
+# Add disable system checks option
```
ekohl: This looks like a copy-paste error.

hooks/pre_validations/13-katello_proxy_reset.rb
```diff
@@ -0,0 +1,16 @@
...
+
+if app_value(:katello_proxy_reset) && !app_value(:noop)
+
+  if param('katello', 'proxy_url').value && param('katello', 'proxy_password').value
+    Kafo::Helpers.log_and_say :info, 'Resetting Katello proxy params...'
+    Kafo::Helpers.reset_value(param('katello', 'proxy_url'))
+    Kafo::Helpers.reset_value(param('katello', 'proxy_port'))
+    Kafo::Helpers.reset_value(param('katello', 'proxy_username'))
+    Kafo::Helpers.reset_value(param('katello', 'proxy_password'))
+  else fail_and_exit('There is no proxy set, exiting')
```
ekohl: Is this really an error? If it's already not set then the desired state has been reached.

chris1984:
@ekohl can I get another review please? test failure is unrelated.

ekohl:
I'm still unsure about this change. Users can already use ```--reset-katello-proxy-url --reset-katello-proxy-port --reset-katello-proxy-username --reset-katello-proxy-password```. It does look like we don't document you can reset answers which is a problem in itself but we already have too many hooks IMHO and I'd like to avoid adding more if we don't really have to.

chris1984:
well this is a result of this bz https://bugzilla.redhat.com/show_bug.cgi?id=1255247

hooks/pre_validations/13-katello_proxy_reset.rb
```diff
@@ -0,0 +1,16 @@
...
+
+if app_value(:katello_proxy_reset) && !app_value(:noop)
+
+  if param('katello', 'proxy_url').value && param('katello', 'proxy_password').value
+    Kafo::Helpers.log_and_say :info, 'Resetting Katello proxy params...'
+    Kafo::Helpers.reset_value(param('katello', 'proxy_url'))
+    Kafo::Helpers.reset_value(param('katello', 'proxy_port'))
+    Kafo::Helpers.reset_value(param('katello', 'proxy_username'))
+    Kafo::Helpers.reset_value(param('katello', 'proxy_password'))
+  else message_and_exit('There is no proxy set, exiting')
```
stbenjam: I'm not sure if it makes sense to exit. I'd expect the message to get logged but the installer to proceed.

stbenjam:
> I'm still unsure about this change. Users can already use --reset-katello-proxy-url --reset-katello-proxy-port --reset-katello-proxy-username --reset-katello-proxy-password. It does look like we don't document you can reset answers which is a problem in itself but we already have too many hooks IMHO and I'd like to avoid adding more if we don't really have to.

I do agree we have waaaaaaay too many hooks.  I think a more elegant solution would to be group these parameters in a "Proxy" category in Kafo and have Kafo supply an option to reset a category to a default.  This makes sense in other cases -- e.g. in puppet-foreman, you might want to clear all the e-mail related settings with one flag instead of a million --reset-param1 --reset-param2 things.

chris1984:
@stbenjam do you have an example?

ekohl:
The linked redmine issue didn't refer there, hence my initial question how it related to it.

It looks like we need to include https://github.com/theforeman/kafo#resetting-an-argument in the foreman-installer manpage and ```--help``` because I think it's more of a problem that users can't find it.

stbenjam:
@chris1984 An example for what?

chris1984:
@stbenjam I think a more elegant solution would to be group these parameters in a "Proxy" category in Kafo and have Kafo supply an option to reset a category to a default.

ekohl:
@stbenjam I do agree that's more elegant and we should be thinking about the more general solution rather than a one-off.

Another thing to consider is the side effect that you will also see this option in the foreman_proxy_content scenario while it won't do anything other than confuse users.

stbenjam:
Kafo supports parameter groups in the documentation, you'll notice in --full-help there's things under various categories.  If proxy settings were grouped in a category called "Katello proxy", then you could do --reset-katello-proxy.  Kafo would need changes and my idea isn't fully thought out, just a suggestion to avoid more hooks since this is a general problem.

chris1984:
@stbenjam i don't know how to proceed with kafo since im not that familiar with that area of the code, can you help when you get a chance.

chris1984:
@stbenjam what kind of change would kafo need done to it?

hooks/pre_validations/13-katello_proxy_reset.rb
```diff
@@ -0,0 +1,16 @@
+def message_and_exit(message)
```
ehelms: This also shouldn't need to be redefined here. There are helpers.

chris1984:
After the discussion on the thread, the better solution would be to update kafo to support param groups with a reset, so closing this pr.

PR #539, opened 2017-09-12
Fixes #20907 - Remove pulp oauth configuration before migrate

ehelms:


ehelms:
Merging given current error is a librarian issue and this passes locally.

PR #541, opened 2017-09-26
Fixes in example snippets in README

iNecas:
There were some missing `\` some extra `\` and most importantly,
using `"~"` in the path doesn't expand the `~`, which leads to
failues:

    Parameter certs-tar invalid: ~/capsule.example.com-certs.tar is
    not one of regexes matching /^(([a-zAError during configuration


ekohl:
Thanks!

PR #542, opened 2017-09-27
Fixes #21128 - expand certs_tar in pre_validations

iNecas:
This allows passing the certs_tar as relative path.

Also, fix the check on exitence of the file, as the certs_tar is now
in different module.


iNecas:
This should work both with `capsule-certs-generate` as well as when passing the `--foreman-proxy-content-certs-tar` to the installer. We check on the exitence of the file only when we actually are about to deploy the certs: checking otherwise makes no sense, as if deploy is false, we are probably just generating the tar

hooks/pre_validations/12-check_capsule_tar.rb
```diff
@@ -4,8 +4,12 @@ def error(message)
   kafo.class.exit 101
 end
 
-if param('capsule', 'certs_tar') && (certs_tar = param('capsule', 'certs_tar').value)
-  unless File.file?(certs_tar)
-    error "The certs tar file generated by the server is not present at #{certs_tar}, exiting."
+certs_tar = param('foreman_proxy_certs', 'certs_tar') ||
```
ekohl: The module ```foreman_proxy_certs``` doesn't exist. Did you mean ```foreman_proxy_content```? This also looks like a bonus fix of previously dead code due to the module rename so 👍

iNecas: In capsule certs generated, there is mapping defined `https://github.com/Katello/katello-installer/blob/master/bin/foreman-proxy-certs-generate#L63` therefor it got here as well. Perhaps @ehelms knows why the mapping is there?

ehelms: Looking at it, I can't figure a real reason other than just typoing it during conversion away from capsule.

ehelms:
You can already pass certs_tar as a relative path for the record [1]. The puppet type is simply 'string' which means no validation is done on pathing. This check adds some robustness to try to ensure that the file does exist at the absolute path and thus has some validity but wanted to clarify.
 
[1] https://github.com/Katello/puppet-certs/blob/master/manifests/foreman_proxy_content.pp#L18

hooks/pre_validations/12-check_capsule_tar.rb
```diff
@@ -4,8 +4,12 @@ def error(message)
...
+  param('foreman_proxy_content', 'certs_tar')
+  
+if certs_tar.value
+  certs_tar.value = File.expand_path(certs_tar.value)
```
ehelms: Do you know if `expand_path` handles `~` and thus also solves that issue we've avoided through docs?

iNecas: Yes, it handles `~` as well

iNecas:
The test failures doesn't seem related

ekohl:
> The test failures don't seem related

https://github.com/voxpupuli/librarian/pull/5 should fix that - it's an incompatibility with git 2.14.0 which Travis has installed.

iNecas:
Anything left on me?

ekohl:
Nothing left, thanks!

PR #543, opened 2017-09-27
Fixes #21138 - Generate proxy certs when no initial org is set

mbacovsky:


ehelms:
Looks good, the message should be updated from capsule to proxy

mbacovsky:
@ehelms thanks, updated

ehelms:
Merging given the unrelated error coming from librarian.

PR #545, opened 2017-10-05
Fixes #21217 - Handle empty string proxy parameters

ekohl:
Also fixes the migration tests and moves a lost capsule migration. Wasn't fully sure if I should include those in this PR.


config/katello.migrations/171005163313-empty-proxy-strings.rb
```diff
@@ -0,0 +1,8 @@
+if answers['katello']
+  ['url', 'port', 'username', 'password'].each do |name|
+    key = "proxy_#{name}"
+    if answers['katello'][key] == ''
```
ehelms: Ahh, and so after this point '' will never be a valid string again? Unless I as a user go edit the answer file directly.

ekohl: Currently the katello module refuses to use them so this clears them which used to be the same. If we do start accepting empty strings again we can remove this migration.

ehelms:
Looks good -- mind squashing down to a single commit ?

ehelms:
Looking good, can squash down to just the proper commit?

ekohl:
Updated

PR #546, opened 2017-10-05
Fixes #21222 - Don't log rpm -q output

ehelms:
When running rpm -q we only care about the result, not to log
the result of the execution. This could lead to ERROR output
when no real error is occurring.

### DIFF
--- a/hooks/boot/01-helpers.rb
+++ b/hooks/boot/01-helpers.rb
@@ -10,7 +10,7 @@ class Kafo::Helpers
       mod.enabled?
     end
 
-    def log_and_say(level, message, do_say = true)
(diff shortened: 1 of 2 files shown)


hooks/boot/01-helpers.rb
```diff
@@ -39,17 +39,17 @@ def execute(commands, do_say = true)
       !results.include? false
     end
 
-    def execute_command(command, do_say)
+    def execute_command(command, do_say, do_log)
```
ekohl: How does Ruby handle a function without a default? I.e. will line 37 break or assume false?

ehelms: Good catch, I updated the file

chris1984:
Thanks @ehelms

PR #547, opened 2017-10-06
Refs #21060 - fix proxy success message to have the right url

evgeni:
before:

    * Foreman Proxy is running at :#<Kafo::Param:20255140 @name="ssl_port" @default=8443 @value=9090 @type=integer (between 0 and 65535)>

after:

    * Foreman Proxy is running at https://fp.example.com:8443


chris1984:
Thanks Evgeni!

PR #548, opened 2017-10-09
Fixes #21248 - Properly copy environments in Puppet 4 upgrade

ehelms:


hooks/init/31-upgrade-puppet.rb
```diff
@@ -25,7 +25,7 @@ def remove_puppet_port_httpd
 
 def copy_data
   success = []
-  success << Kafo::Helpers.execute('cp -rfp /etc/puppet/environments /etc/puppetlabs/code/environments') if File.directory?('/etc/puppet/environments')
+  success << Kafo::Helpers.execute('cp -rfp /etc/puppet/environments/* /etc/puppetlabs/code/environments') if File.directory?('/etc/puppet/environments')
```
ekohl: I think this will break if there are no environments.

ehelms: Is it possible to have an installation that lacks any environments?

ekohl: Not sure, but if a user has a separate Puppet CA from the master it could happen. In my experience it's better to assume it's unexpected and be safe.

ehelms: Does the check at the end not protect us from lack of environments?

ekohl: It doesn't protect you from an empty directory. If we can learn anything from b7a1898a046d8734aa2d983a060de28f08a43edc then we should expect the unexpected.

ehelms: Won't an empty directory just copy nothing and no harm done?

ekohl:
```bash
ekohl@donald /tmp $ mkdir old new
ekohl@donald /tmp $ cp -rfp old/* new
cp: cannot stat 'old/*': No such file or directory
ekohl@donald /tmp $ echo $?
1
```

ehelms:
@ekohl updated with a check

PR #553, opened 2017-10-17
Fixes #21365 - Correct success message for upgrades

mbacovsky:
During recent refactoring we've lost post-install handling of
successfull upgrade. Fixing the condition to get correct
message and "success file" back.


mbacovsky:
@ehelms could you take a look?

PR #560, opened 2017-10-26
Refs #21464 - Remove old CP oauth params

sean797:


config/katello.migrations/171026094113-rename-cp-oauth.rb
```diff
@@ -0,0 +1,2 @@
+answers['katello'].delete('oauth_key') if answers['katello']['oauth_key']
+answers['katello'].delete('oauth_secret') if answers['katello']['oauth_secret']
```
ehelms: This is an edge case, but this should perhaps be `.key?('oauth_secret')` checks, rather than checks on the values. If someone, one of these was set to a falsey value, this would fail.

sean797:
Nice @ehelms 👍 Updated.

config/katello.migrations/171026094113-rename-cp-oauth.rb
```diff
@@ -0,0 +1,2 @@
+answers['katello'].delete('oauth_key') if answers['katello'].key?('oauth_key')
```
ekohl: Would it make sense to just use ```if answers['katello']``` since that would also work if the value is false rather than a hash.

ehelms: It should, I lean towards being specific as possible in general.

ehelms: @ekohl Thoughts? if you feel strongly about your edge case, I am fine with that change being made. Otherwise lets ship it.

ekohl: I do think that having more robust code is a good thing and establishes a common pattern but don't feel strongly about this because I also see reasons for this particular case. Let's just merge this as is.

PR #562, opened 2017-10-27
make single server chef plugin use possible

dschlenk:
fixes #21498 - Chef smart proxy plugin not present in katello scenario's answer file


ehelms:
Thanks @dschlenk -- could you amend your commit and update it with the mssage in your comment `fixes #21498 - Chef smart proxy plugin not present in katello scenario's answer file` ? Then it will be good to go.

PR #563, opened 2017-10-30
fixes #21498 - add Chef proxy plugin to katello's answer file

dschlenk:


ehelms:
Cheers @dschlenk -- thanks for the PR. I'll merge once everything is green,

ehelms:
Although, I typed too fast. This will make this available on new installs only. If you want this option available on upgrades, it will need a migration similar to:

https://github.com/Katello/katello-installer/blob/master/config/katello.migrations/170104152429-add-discovery.rb

There is a script in `bin/` to generate the skeleton for the migration for you.

dschlenk:
thanks @ehelms will do shortly

PR #564, opened 2017-10-30
Fixes #19720 - update foreman-proxy config

chris1984:


config/katello-answers.yaml
```diff
@@ -65,7 +65,7 @@ foreman::plugin::puppetdb: false
 foreman::plugin::remote_execution: false
 foreman::plugin::setup: false
 foreman::plugin::tasks: true
-foreman::plugin::templates: false
+foreman::plugin::templates: true
```
ehelms: This is the Foreman templates plugin, but the Redmine issue mention `--foreman-proxy-templates` which is the smart proxy plugin.

config/katello.migrations/171030135054-foreman-proxy-update.rb
```diff
@@ -0,0 +1,5 @@
+# foreman_proxy defaults
+answers['foreman_proxy'] = {
+  'http' => true,
+  'http_port' => '8000'
```
ehelms: These would also need to be configured in the scenario answers file for fresh install to get them.

config/katello.migrations/171030135054-foreman-proxy-update.rb
```diff
@@ -0,0 +1,6 @@
+# foreman_proxy defaults
+answers['foreman_proxy'] = {
+  'http' => true,
+  'http_port' => '8000',
+  'templates' => true
+}
```
ehelms: This will overwrite everything in `answers['foreman_proxy']` -- note how you are saying take this key in the `answers` hash and set it equal to the hash on the right side. Instead you want to update the individual values within the hash:

`answers['foreman_proxy']['http'] = true if answers['foreman_proxy'].key?('http')`

config/katello-answers.yaml
```diff
@@ -71,6 +72,7 @@ foreman_proxy::plugin::chef: false
 foreman_proxy::plugin::dhcp::infoblox: false
 foreman_proxy::plugin::dns::infoblox: false
 foreman_proxy::plugin::openscap: false
+foreman_proxy::plugin::templates: true
```
ehelms: I think this part can be dropped?

config/katello.migrations/171030135054-foreman-proxy-update.rb
```diff
@@ -0,0 +1,5 @@
+# foreman_proxy defaults for templates
+
+answers['foreman_proxy']['http'] = true if answers['foreman_proxy'].key?('http')
+answers['foreman_proxy']['http_port'] = '8000' if answers['foreman_proxy'].key?('http_port')
```
ehelms: I think this needs to be an unless. I believe the test is failing, because at this point, the answers file has not been "fully" expanded, thus when it evaluates whether or not answers['foreman_proxy'] has 'http_port' as a key, it fails because it's not in the pre-changed answer file.

chris1984:
@ehelms can i get another review please?

ehelms:
LGTM -- @lzap would you mind giving this a once over that it meets your expectations for what a default install configuration should look like?

config/katello.migrations/171030135054-foreman-proxy-update.rb
```diff
@@ -0,0 +1,5 @@
+# foreman_proxy defaults for templates
+
+answers['foreman_proxy']['http'] = true if answers['foreman_proxy'].key?('http')
+answers['foreman_proxy']['http_port'] = '8000' unless answers['foreman_proxy'].key?('http_port')
```
ekohl: The default is an integer ```8000```. This converts it to a string but doesn't really add anything so can be removed.

config/katello-answers.yaml
```diff
@@ -42,9 +42,10 @@ puppet:
   server_foreman_ssl_key: /etc/pki/katello/puppet/puppet_client.key
 foreman_proxy:
   custom_repo: true
-  http: false
+  http: true
+  http_port: '8000'
```
ekohl: The default is an integer ```8000```. I'd expect this to fail but maybe hiera converts it to an integer. Either way, this can be removed.

lzap:
This would match what I expect to be set, thank you.

chris1984:
@ekohl removed the line, can I get review/merge please?

config/katello.migrations/171030135054-foreman-proxy-update.rb
```diff
@@ -0,0 +1,4 @@
+# foreman_proxy defaults for templates
+
+answers['foreman_proxy']['http'] = true if answers['foreman_proxy'].key?('http')
```
ekohl: I think this can simply be ```answers['foreman_proxy']['http'] = true if answers['foreman_proxy']```. In theory ```answers['foreman_proxy']``` can be false or the key could not exist which would make the migration fail/not achieve what you want. It's a little more robust and we should prepare for the unexpected. The same applies below.

PR #570, opened 2017-12-04
Katello 3.5 and cockpit

peppos:
Change default port ssl from 9090 to 9093 because is not compatible with cockpit
https://github.com/theforeman/foreman_cockpit/issues/14


ekohl:
Technically this would also need a migration but before we get into the technical details, it's good to think about the longer term goal.

Vanilla foreman uses port 8443 but in the katello context RHSM also uses 8443. That means we have a conflict. I recall @stbenjam was working on adding a proxy for the RHSM stuff so it could run on 8443 again. I'd prefer that over port 9093 but it'd need some work.

config/foreman-proxy-content.migrations/160120212700-foreman-proxy.rb
```diff
@@ -29,7 +29,7 @@ def mod(manifest, params)
 answers['foreman_proxy'] = {
   'custom_repo' => true,
   'http' => true,
-  'ssl_port' => '9090',
```
ekohl: Migrations are executed only once and then it's stored that it was executed. That means you have to create new migrations. There's a helper script ```bin/create-migration```

peppos:
I have found this problem. When I install Katello and I want to use a cockpit plugin I have a conflit on a TCP port 9090. The 9090 port is used for communication with the Smart Proxy and also the cockpit. I have propose 9093 but another port is valid.

peppos:
My target is: change the default port on answer file when I use the scenario katello.

ekohl:
The bot wants you to modify the commit message. It should be sufficient to squash it all in a single commit and have it include ```Fixes #21859 - ```.

peppos:
Ok! sorry @ekohl

ehelms:
I'm sorry but I think this constitutes a major change whose impact is not
worth it at this time. I would recommend either changing the cockpit port
or customizing this port using the installer in your own installation. I do
not think that cockpits use of the same port warrants such a large change
in our project given cockpit is configurable.

On Dec 7, 2017 4:37 AM, "The Foreman Bot" <notifications@github.com> wrote:

There were the following issues with the commit message:

   - commit message for b60cbad
   <https://github.com/Katello/katello-installer/commit/b60cbad80f009bce82e20127c7c84a9c1e752b4b>
   is not wrapped at 72nd column
   - commit message for 74d4dc3
   <https://github.com/Katello/katello-installer/commit/74d4dc3805bb0f4ff326a8a527e7ecd0bac61054>
   is not wrapped at 72nd column
   - commit message for a8ea84d
   <https://github.com/Katello/katello-installer/commit/a8ea84d6ef097558860a932b5d3518569d1198c8>
   is not wrapped at 72nd column
   - commit message for c280837
   <https://github.com/Katello/katello-installer/commit/c2808376023028f39da147558f43866e090e4d40>
   is not wrapped at 72nd column

If you don't have a ticket number, please create an issue in Redmine
<http://projects.theforeman.org/projects/katello/issues/new>.

More guidelines are available in Coding Standards
<http://theforeman.org/handbook.html#Codingstandards> or on the Foreman wiki
<http://projects.theforeman.org/projects/foreman/wiki/Reviewing_patches-commit_message_format>
.
------------------------------

This message was auto-generated by Foreman's prprocessor
<http://projects.theforeman.org/projects/foreman/wiki/PrProcessor>

—

You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/Katello/katello-installer/pull/570#issuecomment-349913756>,
or mute the thread

ekohl:
I partly agree with @ehelms that changing it to another port is not worth the effort. If we could change our stack back to port 8443 to match vanilla foreman that would be an interesting option. We would need to take care of RHSM in that case. That is a big change though.

peppos:
Yes is the big change. This is valid only if I install katello and not if I install Foreman. If I already use the cockpit in all my servers I must modify all cockpit port and all firewall rules. Alternatively is possible to add a specific option or add the entry in the help like "If you want use the cockpit modify this port" or add this issue in the official documentation. In the mean time I can propose to change the port on cockpit prj if you want.

peppos:
I forgot, if I change the cockpit port I must modify the foreman_cockpit plugin for use a new port.

ekohl:
It's good to note that these are the defaults. You're still free to change the values in your installation by providing these as command line arguments to the installer. See ```--help``` or ```--full-help``` in case they're under advanced sections.

peppos:
I understood that is a default value. But this default is not compatible with default cockpit plugin installation and this is not configurable via option in my installation command but only editing the file.

ekohl:
While I appreciate that, it's simply too big of a change at this point to manage. A lot of our documentation, downstream usage and other integrations would need to be changed. When we do invest time into this, it'd make more sense to align it back with foreman to 8443. I recall @stbenjam having looked into that so maybe he can give his opinion on it.

stbenjam:
>  I recall @stbenjam having looked into that so maybe he can give his opinion on it.

I was only looking at getting rid of using 8443 on the Katello capsule as a reverse proxy.  I think that would need to be done first before we consider moving the proxy port back to 8443 to align with Foreman.

peppos:
Ok, if you want I can test it.

ekohl:
For now it'll need quite a bit of effort and it's not high on my agenda right now. Maybe @stbenjam could outline the needed steps for anyone who wants to work on this.

stbenjam:
For your own environment if you wanted to run the installer with `--foreman-proxy-ssl-port=9093` I'd expect it to work - if you have SELinux permissive or add the appropriate rules to allow Rails to connect to 9093.

The best solution long term would for us to move to 8443, but it's in use by something else.  That needs to get moved to 443, but there needs to be some ability for at least some period of time to support both ports - that's the problem with moving things.  There are many many many client systems out there that connect to 8443 for the subscription-manager service, they need to get moved to 443 but we need to give people _at least_ one release to do it, maybe two.

I started some work over [here](https://github.com/theforeman/puppet-foreman_proxy_content/pull/113) last year, but didn't get a chance to finish it.

ehelms:
@peppos I wanted to clarify something as I think about this. This *only* affects cockpit running on the actual Foreman server with Katello? That is to say, our choice of port currently does not break Cockpit on all the hosts being managed just simply viewing cockpit for the server itself?

peppos:
@ehelms not only for cockpit on Foreman but for all cockpit, because the url is the same for all host. Only the hostname is a variable in the plugin code. From my point of view the are three ways, modify the port for all cockpit services on all servers or modify the foreman proxy port or modify the plugin code and assume that port is a variable.

ehelms:
I think option 3 has the most flexibility and would require the least amount of re-configuration in user environments. Further, that allows an organization to modify the cockpit ports across hosts if they have requirements around port usage.

ekohl:

PR #573, opened 2018-01-08
fixes #22173 - clear msg when fwd ip doesn't resolve

stbenjam:
When an IP address is included in the forward resolution, but has no
reverse record at all, we throw an exception instead of showing an
error message.

Before:
```
/usr/share/ruby/resolv.rb:130:in `getname': no name for 192.168.1.6
(Resolv::ResolvError) from /usr/share/ruby/resolv.rb:66:in `getname'
        from ./dns.rb:27:in `block in <main>' from ./dns.rb:21:in `each'
        from ./dns.rb:21:in `<main>'
```

After:
```
Forward DNS 192.168.1.6 did not reverse resolve to any hostname.
```

### DIFF
--- a/checks/dns.rb
+++ b/checks/dns.rb
@@ -24,8 +24,12 @@ forwards.each do |ip|
     error_exit("Forward DNS points to #{ip} which is not configured on this server")
   end
 
-  reverse = Resolv.getname(ip.to_s)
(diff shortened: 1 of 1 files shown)


stbenjam:
Thanks!

PR #575, opened 2018-01-10
fixes #22230 - reindex docker tags on upgrade

thomasmckay:
Relies on https://github.com/Katello/katello/pull/7142


jlsherrill:
@ekohl i think that makes sense for a 3.7 feature!

PR #578, opened 2018-01-11
Refs #21703 - remove registration tasks

beav:
`Host::Register` tasks have been removed as part of
https://github.com/Katello/katello/pull/7082.

This change adds a step to remove all tasks with label
`Actions::Katello::Host::Register`. Without this, already-executed tasks will
show as blank in the web UI.


beav:
do not merge until https://github.com/Katello/katello/pull/7082 is merged.

PR #579, opened 2018-01-16
refs #22173 - handle ipv6 on broken rubies

stbenjam:
Ruby < 2.3 doesn't support getname on IPv6 addresses: https://bugs.ruby-lang.org/issues/12112

My first thought was to just skip the test on old ruby:
```
next if ip.ipv6? && Gem::Version.new(RUBY_VERSION) < Gem::Version.new('2.3.0')
```
@ekohl proposed using getresources, which works in my test environment that has real DNS servers. However, if a user expects /etc/hosts to work for forward/reverse lookups, it won't with getresources.  

Only Resolv::Hosts seems to do that, and that's broken with IPv6.

Thoughts?

### DIFF
--- a/checks/dns.rb
+++ b/checks/dns.rb
@@ -30,6 +30,14 @@ forwards.each do |ip|
       error_exit("Reverse DNS #{reverse} does not match hostname #{hostname}")
     end
   rescue Resolv::ResolvError
+    # Because of https://bugs.ruby-lang.org/issues/12112:
+    if ip.ipv6?
+      reverse = Resolv::DNS.open do |dns|
+        dns.getresources ip.ip6_arpa, Resolv::DNS::Resource::IN::PTR
+      end.first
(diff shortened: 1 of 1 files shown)


ekohl:
I think not using /etc/hosts could be a feature because reverse DNS also matters for other hosts - not just the single one but how about a fallback? Always do getname (for modern Rubies) and only use this code if the reverse was not found and it's IPv6.

stbenjam:
> I think not using /etc/hosts could be a feature sometimes 

Yea, sometimes, but there's some really broken environments where /etc/hosts is just kind of used everywhere even on clients.  Hopefully anyone who's ahead of the times and deploying IPv6 everywhere has working DNS, at least until Ruby 2.3 is standard or a fix is backported.

> how about a fallback?

I like that idea - how about this?

stbenjam:
Thanks all

PR #581, opened 2018-01-26
Fixes #22431 - prevent migration of custom hammer credentials

mbacovsky:
Cases I focused on are when the credentials are manually altered i.e. username is not the default 'admin' or the password is set. In such cases I presume the user has hammer configured and is using it so I'm preventing the installer to touch it.


mbacovsky:
The CI failure seems unrelated.

mbacovsky:
Related fix in the foreman-installer https://github.com/theforeman/foreman-installer/pull/260

ekohl:
Thanks!

PR #583, opened 2018-01-29
Pin modules to 3.6 versions

ekohl:
Needs https://github.com/theforeman/foreman-installer/pull/261 for a Puppetfile.lock


ehelms:
I don't understand what's causing the build error.

I think we should start pinning this as well in release branches -- https://github.com/Katello/katello-installer/blob/master/Rakefile#L16

ekohl:
Should we set the env for Travis or change the Rakefile?

ekohl:
Updated to remove the katello-devel scenario.

PR #587, opened 2018-02-12
Fixes #20054 - Add test if server's certificate is PEM encoded.

phayz:


chris1984:
now that is rebased, ill test with my der format certs

PR #588, opened 2018-02-12
Update puppet modules for 3.5.1

akofink:


Puppetfile
```diff
@@ -3,13 +3,10 @@ forge 'https://forgeapi.puppetlabs.com'
...
+mod 'katello/qpid',                  '>= 4.0.0 < 4.1.0'
+mod 'katello/pulp',                  '>= 5.1.0 < 5.2.0'
+mod 'katello/candlepin',             '>= 5.0.0 < 5.1.0'
 
-mod 'katello/foreman_proxy_content', '>= 6.0.0 < 6.1.0'
+mod 'katello/foreman_proxy_content', '>= 7.0.0 < 7.1.0'
 mod 'katello/katello',               '>= 7.0.0 < 7.1.0'
-
-# Katello devel specific modules
-mod 'katello/katello_devel',         :git => 'https://github.com/Katello/puppet-katello_devel'
```
ekohl: If we drop this here, you also need some other changes like in https://github.com/Katello/katello-installer/commit/9fbd7a17ad899708193161c8969141971c363d92 (drop answers, remove it from tests, Rakefile)

Puppetfile
```diff
@@ -3,13 +3,10 @@ forge 'https://forgeapi.puppetlabs.com'
...
+mod 'katello/pulp',                  '>= 5.3.0 < 5.4.0'
+mod 'katello/candlepin',             '>= 5.0.0 < 5.1.0'
 
-mod 'katello/foreman_proxy_content', '>= 6.0.0 < 6.1.0'
-mod 'katello/katello',               '>= 7.0.0 < 7.1.0'
-
-# Katello devel specific modules
-mod 'katello/katello_devel',         :git => 'https://github.com/Katello/puppet-katello_devel'
+mod 'katello/foreman_proxy_content', '>= 7.0.0 < 7.1.0'
+mod 'katello/katello',               '>= 7.2.0 < 7.3.0'
```
akofink: All of the changes in this file are motivated by [this issue](http://projects.theforeman.org/issues/22289), which requires https://github.com/theforeman/puppet-foreman_proxy_content/commit/d1155d5f5405d23164ad0cb13580beb75bf98873 (7.0.0) and https://github.com/theforeman/puppet-qpid/commit/6a0450649716c317437c8e612fda2459fba8dd27 (4.0.0).

Puppetfile.lock
```diff
@@ -19,141 +19,105 @@ FORGE
...
-  katello-katello (< 7.1.0, >= 7.0.0)
-  katello-katello_devel (>= 0)
-  katello-pulp (< 5.1.0, >= 5.0.0)
-  katello-qpid (< 3.2.0, >= 3.1.0)
+  katello-foreman_proxy_content (< 7.1.0, >= 7.0.0)
+  katello-katello (< 7.3.0, >= 7.2.0)
+  katello-pulp (< 5.4.0, >= 5.3.0)
+  katello-qpid (< 4.0.1, >= 4.0.0)
   katello-service_wait (< 1.1.0, >= 1.0.3)
-  puppetlabs-mysql (< 4.1.0, >= 4.0.0)
```
ekohl: You should run the update with the foreman-installer (1.16) Puppetfile appended to this Puppetfile so these are locked to their versions. Now you're creating a mismatch in versions.

Puppetfile
```diff
@@ -3,13 +3,10 @@ forge 'https://forgeapi.puppetlabs.com'
 mod 'katello/common',                '>= 1.0.1 < 1.1.0'
 mod 'katello/service_wait',          '>= 1.0.3 < 1.1.0'
 
-mod 'katello/certs',                 '>= 3.0.0 < 3.1.0'
-mod 'katello/qpid',                  '>= 3.1.0 < 3.2.0'
-mod 'katello/pulp',                  '>= 5.0.0 < 5.1.0'
-mod 'katello/candlepin',             '>= 4.0.0 < 4.1.0'
+mod 'katello/certs',                 '>= 3.0.1 < 3.0.2'
```
ekohl: Why does this pin need changing? It already allowed the update to happen.

Puppetfile
```diff
@@ -3,13 +3,10 @@ forge 'https://forgeapi.puppetlabs.com'
...
+mod 'katello/qpid',                  '>= 4.0.0 < 4.0.1'
```
ekohl: We typically pin on the minor version, not patch level. Patch level pins happen in the lock file. So it should be ```< 4.1.0```.

Puppetfile.lock
```diff
@@ -1,12 +1,12 @@
 FORGE
   remote: https://forgeapi.puppetlabs.com
   specs:
-    camptocamp-systemd (1.0.0)
+    camptocamp-systemd (1.1.1)
```
ekohl: Unless there's an issue I know that should be fixed I usually keep it at the older version for third part releases, just to be safe.

Puppetfile
akofink: Typo, will revert.

Puppetfile
```diff
@@ -3,13 +3,10 @@ forge 'https://forgeapi.puppetlabs.com'
...
+mod 'katello/qpid',                  '>= 4.0.0 < 4.1.0'
+mod 'katello/pulp',                  '>= 5.3.0 < 5.4.0'
+mod 'katello/candlepin',             '>= 5.0.0 < 5.1.0'
 
-mod 'katello/foreman_proxy_content', '>= 6.0.0 < 6.1.0'
-mod 'katello/katello',               '>= 7.0.0 < 7.1.0'
-
-# Katello devel specific modules
-mod 'katello/katello_devel',         :git => 'https://github.com/Katello/puppet-katello_devel'
+mod 'katello/foreman_proxy_content', '>= 7.0.0 < 7.1.0'
```
ehelms: This chnage will pull in a variety of changes:

 1) http://projects.theforeman.org/issues/21464 -- removal of pulp-oauth support, this requires multiple changes
 2) http://projects.theforeman.org/issues/21430 -- this would need to be re-aligned
 3) http://projects.theforeman.org/issues/22338 -- which is only available as of Pulp 2.15 I believe

Thus, I think we are going to have to do a 6.1.z release with the puppet-qpid support for dispatch router

Puppetfile.lock
```diff
@@ -1,12 +1,12 @@
 FORGE
   remote: https://forgeapi.puppetlabs.com
   specs:
-    camptocamp-systemd (1.0.0)
+    camptocamp-systemd (1.1.0)
```
ekohl: Any reason to switch to 1.1.0? I'd either stay with 1.0.0 or go to 1.1.1.

Puppetfile.lock
```diff
@@ -19,66 +19,54 @@ FORGE
...
       camptocamp-systemd (< 2.0.0, >= 0.4.0)
       puppetlabs-concat (< 5.0.0, >= 1.0.0)
       puppetlabs-stdlib (< 5.0.0, >= 4.13.0)
     katello-service_wait (1.0.3)
       puppetlabs-stdlib (< 5.0.0, >= 4.2.0)
     puppet-extlib (1.1.0)
       puppetlabs-stdlib (>= 0)
-    puppet-staging (2.2.0)
-    puppetlabs-apache (2.3.0)
+    puppet-mongodb (2.0.0)
```
ekohl: Not sure about switching from puppetlabs-mongodb 1.x to puppet-mongodb 2.0. Our usage of mongodb parameters is minimal so it's probably not a problem but there is a breaking change in there.

Puppetfile
```diff
@@ -1,15 +1,30 @@
 forge 'https://forgeapi.puppetlabs.com'
 
+# mod 'puppet/mongodb',        '>= 1.0.0 < 2.0.0'
```
ekohl: Could you leave these out? It's just duplication.

Puppetfile
```diff
@@ -1,15 +1,30 @@
...
 
 mod 'katello/certs',                 '>= 3.0.0 < 3.1.0'
-mod 'katello/qpid',                  '>= 3.1.0 < 3.2.0'
-mod 'katello/pulp',                  '>= 5.0.0 < 5.1.0'
-mod 'katello/candlepin',             '>= 4.0.0 < 4.1.0'
-
-mod 'katello/foreman_proxy_content', '>= 6.0.0 < 6.1.0'
-mod 'katello/katello',               '>= 7.0.0 < 7.1.0'
+mod 'katello/qpid',                  '>= 4.0.0 < 4.1.0'
+mod 'katello/pulp',                  '>= 5.3.0 < 5.4.0'
```
ekohl: 5.0.1 has been released.

Puppetfile
```diff
@@ -1,15 +1,30 @@
...
+mod 'katello/candlepin',             '>= 5.0.0 < 5.1.0'
```
ekohl: 4.0.1 has been released.

Puppetfile.lock
```diff
@@ -118,42 +105,13 @@ FORGE
...
-  katello-katello (< 7.1.0, >= 7.0.0)
-  katello-katello_devel (>= 0)
-  katello-pulp (< 5.1.0, >= 5.0.0)
-  katello-qpid (< 3.2.0, >= 3.1.0)
+  katello-foreman_proxy_content (< 6.2.0, >= 6.1.0)
+  katello-katello (< 7.3.0, >= 7.2.0)
+  katello-pulp (< 5.4.0, >= 5.3.0)
+  katello-qpid (< 4.1.0, >= 4.0.0)
   katello-service_wait (< 1.1.0, >= 1.0.3)
-  puppetlabs-mysql (< 4.1.0, >= 4.0.0)
```
ekohl: Could you run the update with these modules in the Puppetfile so they stay locked on the right versions?

Puppetfile.lock
```diff
@@ -72,18 +71,16 @@ FORGE
...
+      puppetlabs-inifile (< 3.0.0, >= 1.1.3)
       puppetlabs-postgresql (< 6.0.0, >= 4.0.0)
       puppetlabs-stdlib (< 5.0.0, >= 4.2.2)
     puppetlabs-stdlib (4.19.0)
-    puppetlabs-vcsrepo (2.0.0)
     puppetlabs-xinetd (3.0.0)
       puppetlabs-stdlib (< 5.0.0, >= 4.6.0)
     richardc-datacat (0.6.2)
     theforeman-dhcp (4.0.2)
-      camptocamp-systemd (< 2.0.0, >= 1.0.0)
```
ekohl: This is odd because 4.0.2 does depend on camptocamp-systemd

Puppetfile.lock
```diff
@@ -19,41 +19,40 @@ FORGE
...
+    katello-qpid (4.0.0)
       puppetlabs-concat (< 5.0.0, >= 1.0.0)
       puppetlabs-stdlib (< 5.0.0, >= 4.13.0)
     katello-service_wait (1.0.3)
       puppetlabs-stdlib (< 5.0.0, >= 4.2.0)
     puppet-extlib (1.1.0)
       puppetlabs-stdlib (>= 0)
     puppet-staging (2.2.0)
-    puppetlabs-apache (2.3.0)
+    puppetlabs-apache (2.3.1)
```
ekohl: This is sort of a lie since it's actually shipped in foreman-installer which only has 2.3.0. In the past we've seen weird issues with cache mismatches. It'd be best to keep this change out.

PR #589, opened 2018-02-12
Fixes #22558 - Add post install devel instructions

ekohl:


hooks/boot/02-message-helpers.rb
```diff
@@ -9,6 +9,14 @@ def server_success_message(kafo)
       say "  * <%= color('Katello', :info) %> is running at <%= color('#{kafo.param('foreman', 'foreman_url').value}', :info) %>"
     end
 
+    def dev_server_success_message(kafo)
+      success_message
+      say "  * <%= color('Katello', :info) %> is running at <%= color('https://#{`hostname -f`}', :info) %>"
```
ehelms: This is a bit of a false claim right? There is no actual server running at the time of installation completion. I think this is a good general idea, but I'd prefer to re-word it for a bit more accuracy and not set any wrong expectations for users.

ekohl: You're right, it should tell the user to log in, run foreman and then it should be running there.

ekohl: Updated

PR #591, opened 2018-02-15
Fixes #22598 - add san check

chris1984:


bin/katello-certs-check
```diff
@@ -112,11 +112,22 @@ function check-ca-bundle () {
...
 }
 
+function check-cert-san () {
+    printf "Checking Subject Alt Name on certificate"
+    CHECK=$(openssl x509 -noout -text -in $CERT_FILE | grep DNS:)
+    if [ $? == "0" ]; then
+        success
+  else
+        error 4 "The $CERT_FILE does not contain a Subject Alt Name"
+  fi
```
ehelms: Indentation is off on the else/if

PR #595, opened 2018-02-27
Fixes #22703 - Look for /var/lib/puppet first

upadhyeammit:
This would first look for /var/lib/puppet directory for candlepin_db_password file instead of going in 
/opt/puppetlabs/puppet/cache/ directory, this ensures if puppetlab's puppet agent has installed then installer should not fail to get file candlepin_db_password.

### DIFF
--- a/hooks/boot/01-helpers.rb
+++ b/hooks/boot/01-helpers.rb
@@ -1,7 +1,7 @@
 class Kafo::Helpers
   class << self
     def puppet_dir
-      @puppet_dir ||= File.directory?('/opt/puppetlabs/puppet') ? '/opt/puppetlabs/puppet/cache' : '/var/lib/puppet'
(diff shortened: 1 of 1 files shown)


ekohl:
Upstream has dropped support for Puppet 3 so almost of all deployments should be Puppet 4 or Puppet 5 AIO installs in ```/opt/puppetlabs/puppet```. Only Debian with non-AIO packages still has ```/var/lib/puppet``` but katello-installer doesn't suppport Debian. Current Kafo versions also would prefer a Puppet Enterprise over Puppet 3 if both are installed. That means this patch is very likely to break for all our current users and should be a downstream 6.2 only fix given 6.3 supports Puppet 4 in the same way upstream does.

ekohl:
Closing because this is not a good idea. It moves in the opposite direction we want to move.

PR #599, opened 2018-03-01
Refs #22558: Add a nil check on foreman auth

akofink:
@ekohl This is nil for me. I'm not sure if that's an expected case. Related: https://github.com/Katello/katello-installer/pull/589

```
[ INFO 2018-03-01T18:59:59 verbose]  Katello_devel::Bundle[exec rails s -d]: Scheduling refresh of Exec[bundle-exec rails s -d]                                                                                                               
[ WARN 2018-03-01T19:00:14 verbose]  /Stage[main]/Katello_devel::Setup/Katello_devel::Bundle[exec rails s -d]/Exec[bundle-exec rails s -d]/returns: executed successfully                                                                     
[ WARN 2018-03-01T19:00:14 verbose]  /Stage[main]/Katello_devel::Setup/Katello_devel::Bundle[exec rails s -d]/Exec[bundle-exec rails s -d]: Triggered 'refresh' from 1 events                                                                 
[ WARN 2018-03-01T19:00:15 verbose]  /Stage[main]/Katello_devel::Setup/Exec[destroy rails server]/returns: executed successfully                                                                                                              
[ WARN 2018-03-01T19:00:15 verbose]  /Stage[main]/Katello_devel::Setup/Exec[destroy rails server]: Triggered 'refresh' from 1 events                                                                                                          
[ WARN 2018-03-01T19:00:15 verbose]  Applied catalog in 513.60 seconds                                                 
[ INFO 2018-03-01T19:00:17 verbose] Puppet has finished, bye!                                                          
[ INFO 2018-03-01T19:00:17 verbose] Executing hooks in group post                                                      
  Success!                                                 
  * To run the Katello dev server log in using SSH         
  * Run `cd foreman && bundle exec foreman start`          
  * The server is running at https://centos7.akofink-desktop                                                           
                                                           
  * On Firefox you need to accept the certificate at https://centos7.akofink-desktop                                   
:3808                                                                                                                  
[ INFO 2018-03-01T19:00:17 verbose] Installer finished in 525.71697479 seconds                                         
                                                                                                                                                                                                                                              
                                                                                                                       
STDERR:                                                    
                                                                                                                       
/usr/share/katello-installer-base/hooks/post/10-post_install.rb:21:in `block (4 levels) in load': undefined method `value' for nil:NilClass (NoMethodError)                                                                                   
        from /usr/share/gems/gems/kafo-2.1.0/lib/kafo/hooking.rb:34:in `instance_eval'                                                                                                                                                        
[...]

PR #600, opened 2018-03-06
Refs #22558: Handle the devel credentials

ekohl:
The devel scenario doesn't use the foreman module so this always fails.

Replaces https://github.com/Katello/katello-installer/pull/599


hooks/boot/02-message-helpers.rb
```diff
@@ -37,6 +37,10 @@ def new_install_message(kafo)
       say "      Initial credentials are <%= color('#{kafo.param('foreman', 'admin_username').value}', :info) %> / <%= color('#{kafo.param('foreman', 'admin_password').value}', :info) %>"
     end
 
+    def dev_new_install_message(kafo)
+      say "      Initial credentials are <%= color('admin', :info) %> / <%= color('#{kafo.param('katello_devel', 'admin_password').value}', :info) %>"
```
akofink: very nice :+1:

PR #602, opened 2018-03-14
Fixes #22905 - Remove katello_devel and devel scenarios

ehelms:


ekohl:
I realized this breaks because we depend on the puppetlabs-vcsrepo module. We need some way to get this dependency.

ehelms:
We need it for the puppet-katello_devel in the new Forklift work or we need it for modules that aren't puppet-katello_devel ?

ekohl:
I believe it's only required by the devel module if you remove it here, you need to install it in forklift

ehelms:
Fix here -- https://github.com/theforeman/forklift/pull/659

ehelms:
@ekohl Given https://github.com/theforeman/forklift/pull/659 is merged, can you re-review?

PR #607, opened 2018-04-05
Refs #23025 escape db connection string

alda519:


alda519:
See also https://github.com/theforeman/puppet-candlepin/pull/101

Not sure if it is better to fix it in candlepin's cpdb code to quote parameters.

alda519:
removed ```\\``` from

     db_uri += "\\&sslfactory=org.postgresql.ssl.NonValidatingFactory" unless db_ssl_verify

this should be handled by cpdb

and added quotes to username and password

ekohl:
Thanks!

PR #608, opened 2018-04-05
Add a script to pin dependencies during release

ekohl:


Rakefile
```diff
@@ -152,3 +160,36 @@ CLEAN.include(BUILDDIR)
 CLEAN.include(PKGDIR)
 
 task :default => [:rubocop, :spec, 'pkg:generate_source']
+
+if pin_task
+  desc 'Pin all the modules in Puppetfile to released versions instead of git branches'
+  task :pin_modules do
+    filename = 'Puppetfile'
+    $new_content = []
```
stbenjam: Why a global variable?

ekohl: Basically I load `Puppetfile` as a ruby script and handle `forge` and `mod` as ruby functions. Because I want to buffer the output somewhere and can't modify the Puppetfile to store it somewhere, this is the only way I knew how to do this. If there's a better way, I'd love to hear it.

stbenjam: Oh I see - clever.

ekohl:
> Works well, just need to turn off the rubocop check for globals.

Rubocop is happy

ehelms:
Just needs an issue attached

PR #609, opened 2018-04-09
Refs #23030 - Add upgrade step for Mongo 3.x

chris1984:


hooks/pre/30-upgrade.rb
```diff
@@ -80,6 +80,22 @@ def fix_katello_settings_file
   end
 end
 
+def mongo_mmapv1_check
+  custom_hiera = '/etc/foreman-installer/custom-hiera.yaml'
+  # check and see if we have a pulp_database already and if the Hiera file has already been modifed.
+  if File.file?('/var/lib/mongodb/pulp_database.0')
+    if File.foreach("#{custom_hiera}").grep(/mongodb::server::storage_engine:/).any?
```
ekohl: Would it make sense to just parse the YAML? Not sure if the ruby yaml parser maintains comments though.

hooks/pre/30-upgrade.rb
```diff
@@ -80,6 +80,22 @@ def fix_katello_settings_file
...
+      logger.info 'No changed needed, Mongo storage engine will be set to Wired Tiger'
+    else
+      # set storage engine to MMAPv1 in Hiera file.
+      logger.info 'Detecting Pulp database, keeping storage engine as MMAPv1'
+      logger.info 'To upgrade to wiredTiger at a later time run foreman-installer with the --upgrade-mongo-storage flag.'
+      `echo '# Added by foreman-installer during upgrade, do not remove or Mongo will not start' >> #{custom_hiera}`
```
ekohl: I'd prefer to use native ruby to append to the file.

hooks/pre/30-upgrade.rb
```diff
@@ -80,6 +80,22 @@ def fix_katello_settings_file
   end
 end
 
+def mongo_mmapv1_check
+  custom_hiera = '/etc/foreman-installer/custom-hiera.yaml'
+  # check and see if we have a pulp_database already and if the Hiera file has already been modifed.
+  if File.file?('/var/lib/mongodb/pulp_database.0')
```
ekohl: Is this file always an indication of `mmapv1` or is it also present with `wiredTiger`?

hooks/pre/30-upgrade.rb
```diff
@@ -80,6 +80,22 @@ def fix_katello_settings_file
   end
 end
 
+def mongo_mmapv1_check
+  custom_hiera = '/etc/foreman-installer/custom-hiera.yaml'
```
ekohl: Hardcoding this path is generally not a good idea, but I guess it already happens a lot.

hooks/pre/30-upgrade.rb
```diff
@@ -45,12 +45,12 @@ def remove_gutterball
 def migrate_pulp
   # Fix pid if neccessary
   if Kafo::Helpers.execute("grep -qe '7.[[:digit:]]' /etc/redhat-release")
-    Kafo::Helpers.execute("sed -i -e 's?/var/run/mongodb/mongodb.pid?/var/run/mongodb/mongod.pid?g' /etc/mongod.conf")
+    Kafo::Helpers.execute("sed -i -e 's?/var/opt/rh/rh-mongodb34/run/mongodb/mongod.pid?/var/opt/rh/rh-mongodb34/run/mongodb/mongod.pid?g' /etc/opt/rh/rh-mongodb34/mongod.conf")
```
ekohl: Am I missing something or is this a noop? I think this whole block can be deleted since can assume the pid is just fine in the SCL.

hooks/pre/30-upgrade.rb
```diff
@@ -4,11 +4,11 @@
 SSL_BUILD_DIR = param('certs', 'ssl_build_dir').value
 
 def stop_services
-  Kafo::Helpers.execute('katello-service stop --exclude mongod,postgresql')
+  Kafo::Helpers.execute('katello-service stop --exclude rh-mongodb34-mongod,postgresql')
```
ekohl: Should we still make sure the old mongo is stopped before starting?

hooks/pre/30-upgrade.rb
```diff
@@ -45,12 +45,12 @@ def remove_gutterball
 def migrate_pulp
   # Fix pid if neccessary
   if Kafo::Helpers.execute("grep -qe '7.[[:digit:]]' /etc/redhat-release")
-    Kafo::Helpers.execute("sed -i -e 's?/var/run/mongodb/mongodb.pid?/var/run/mongodb/mongod.pid?g' /etc/mongod.conf")
+    Kafo::Helpers.execute("sed -i -e 's?/var/opt/rh/rh-mongodb34/run/mongodb/mongod.pid?/var/opt/rh/rh-mongodb34/run/mongodb/mongod.pid?g' /etc/opt/rh/rh-mongodb34/mongod.conf")
   end
 
   # Start mongo if not running
```
ekohl: I wonder if we should just remove this block and the `pulp-manage-db`. If you remove `/var/lib/pulp/init.flag` then the installer will just migrate the DB again and ensure the service is running.

It may be needed to verify the `remove_node_importers` and `remove_node_distributors` can still run though.

hooks/pre/10-reset_feature.rb
```diff
@@ -48,9 +48,9 @@ def reset_pulp
   commands = [
     'rm -f /var/lib/pulp/init.flag',
     'service-wait httpd stop',
-    'service-wait mongod stop',
+    'service-wait rh-mongodb34-mongod stop',
```
ekohl: I wonder if this should also ensure the old mongo is stopped.

hooks/pre/30-upgrade.rb
chris1984: @ekohl it would be present with `wiredTiger` too

hooks/pre/30-upgrade.rb
chris1984: :crying_cat_face:

hooks/pre/30-upgrade.rb
ekohl: Is there a way to detect the database format? If this hook is running on a system that for some reason is already wiredTiger (database restore?) it'll break. It'd be great if we could detect that.

hooks/pre/30-upgrade.rb
ehelms: I agree, I think this was an old work around we shouldnt need anymore.

hooks/pre/30-upgrade.rb
ehelms: As long as we don't write the file, reading it as parsed yaml should be fine. This seems to be fairly clean though.

hooks/post/30-upgrade.rb
```diff
@@ -137,6 +138,24 @@ def remove_event_queue
   end
 end
 
+def remove_legacy_mongo
+    logger.info 'Verifying if Mongo 3.x is running'
+    # Start mongo if not running
+    unless Kafo::Helpers.execute('pgrep mongod')
+      Kafo::Helpers.execute('service-wait rh-mongodb34-mongod start')
+    end
```
ehelms: Just curious why this check is needed? Given the installer should start it if it reached this point.

hooks/pre/10-reset_feature.rb
```diff
@@ -48,9 +48,10 @@ def reset_pulp
...
+    'systemctl stop mongod > /dev/null 2>&1', # piping to dev null incase the old mongo is not installed so we dont get a traceback.
```
ehelms: What does this solve?

hooks/pre/30-upgrade.rb
```diff
@@ -1,14 +1,17 @@
 require 'fileutils'
+require 'yaml'
 
 STEP_DIRECTORY = '/etc/foreman-installer/applied_hooks/pre/'
 SSL_BUILD_DIR = param('certs', 'ssl_build_dir').value
+MONGO_ENGINE_MMAPV1 = '/etc/foreman-installer/.mongo_engine_mmapv1'.freeze
 
 def stop_services
-  Kafo::Helpers.execute('katello-service stop --exclude mongod,postgresql')
+  Kafo::Helpers.execute('systemctl stop mongod')
```
ehelms: What happens if `mongod` no longer exists?

hooks/pre/30-upgrade.rb
```diff
@@ -80,6 +79,30 @@ def fix_katello_settings_file
   end
 end
 
+def mongo_mmapv1_check
+  custom_hiera = '/etc/foreman-installer/custom-hiera.yaml'
+  mongodb_dir = '/var/lib/mongodb/'
+  # check and see if we have a pulp_database already, we have a wiredTiger file, and if the Hiera file has already been modifed.
+  if File.file?("#{mongodb_dir}/pulp_database.0") && File.file?("#{mongodb_dir}/WiredTiger.wt")
```
ehelms: So if the pulp database exists, and a wiredTiger file does not exist, we skip this block? I think this block was needed to ensure we kept `mmapv1`?

hooks/pre/30-upgrade.rb
```diff
@@ -1,14 +1,17 @@
 require 'fileutils'
+require 'yaml'
```
ehelms: I dont see this used anywhere.

hooks/post/30-upgrade.rb
```diff
@@ -1,6 +1,7 @@
 require 'fileutils'
 
 STEP_DIRECTORY = '/etc/foreman-installer/applied_hooks/post/'
+MONGO_REMOVAL_COMPLETE = '/etc/foreman-installer/.mongo_2_removed'.freeze
```
ekohl: We're starting to get more and more of these markers. I wonder if it's time for a `/var/lib/foreman-installer` for various state things, but let's leave that for a later stage.

hooks/post/30-upgrade.rb
```diff
@@ -137,6 +138,24 @@ def remove_event_queue
...
+def remove_legacy_mongo
+    logger.info 'Verifying if Mongo 3.x is running'
+    # Start mongo if not running
+    unless Kafo::Helpers.execute('pgrep mongod')
+      Kafo::Helpers.execute('service-wait rh-mongodb34-mongod start')
+    end
+    # Check to see if the RPMS exist and if so remove them and create the upgrade done file.
+    if `rpm -q mongodb --queryformat=%{version}`.start_with?('2.')
+      logger.warn 'removing Mongo 2.x packages'
+      Kafo::Helpers.execute("yum remove -y mongodb-2* mongodb-server-2* /dev/null 2>&1")
```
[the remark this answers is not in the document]
ekohl: There's also a redirect to /dev/null missing. I think that calling `execute(cmd, false)` will also silence it but preserve it in the logs which is cleaner.

hooks/pre/30-upgrade.rb
```diff
@@ -80,6 +79,30 @@ def fix_katello_settings_file
...
+    else
+      # set storage engine to MMAPv1 in Hiera file and create engine file.
+      logger.info 'Detecting Pulp database and no wiredTiger files, keeping storage engine as MMAPv1'
+      logger.info 'To upgrade to wiredTiger at a later time run foreman-installer with the --upgrade-mongo-storage flag.'
+      # Write to custom_hiera to tell users to not touch the setting.
+      open(custom_hiera, 'a') { |f|
+        f << "# Added by foreman-installer during upgrade, do not remove or Mongo will not start!!!!\n"
+        f << "mongodb::server::storage_engine: 'mmapv1'\n"
+      }
+      # Create engine file so we know Mongo is in mmapv1 for engine upgrade hook.
```
ekohl: We can detect that ourselves at runtime so I'd prefer to not introduce another marker. Markers are used when you can't detect what happened.

hooks/pre/30-upgrade.rb
```diff
@@ -80,6 +79,30 @@ def fix_katello_settings_file
...
+  if File.file?("#{mongodb_dir}/pulp_database.0") && File.file?("#{mongodb_dir}/WiredTiger.wt")
+    if File.foreach("#{custom_hiera}").grep(/mongodb::server::storage_engine:/).any?
+      logger.info 'No changed needed, Mongo storage engine will installed with wiredTiger'
+    else
+      # set storage engine to MMAPv1 in Hiera file and create engine file.
+      logger.info 'Detecting Pulp database and no wiredTiger files, keeping storage engine as MMAPv1'
+      logger.info 'To upgrade to wiredTiger at a later time run foreman-installer with the --upgrade-mongo-storage flag.'
+      # Write to custom_hiera to tell users to not touch the setting.
+      open(custom_hiera, 'a') { |f|
+        f << "# Added by foreman-installer during upgrade, do not remove or Mongo will not start!!!!\n"
```
ekohl: I think the exclamation points are slightly excessive. Perhaps it should instruct the user about the ```upgrade-mongo-storage``` here as well since that's the proper fix.

hooks/pre/30-upgrade.rb
```diff
@@ -43,14 +44,13 @@ def remove_gutterball
...
-    Kafo::Helpers.execute("sed -i -e 's?/var/run/mongodb/mongodb.pid?/var/run/mongodb/mongod.pid?g' /etc/mongod.conf")
-  end
-
   # Start mongo if not running
   unless Kafo::Helpers.execute('pgrep mongod')
-    Kafo::Helpers.execute('service-wait mongod start')
+    if `rpm -q mongodb --queryformat=%{version}`.start_with?('2.') # If mongo 2.x is on the system run the migration with that.
+      Kafo::Helpers.execute('systemctl start mongod')
+    else
+    Kafo::Helpers.execute('service-wait rh-mongodb34-mongod start')
```
ehelms: Indent 2 spaces

ehelms:
These two lines also causes problems right now:

 * https://github.com/Katello/katello-installer/pull/609/files#diff-45212ce99fdfbd0351c83af485751584R244
 * https://github.com/Katello/katello-installer/pull/609/files#diff-45212ce99fdfbd0351c83af485751584R253

I think we may also need to ensure `mongod` is stopped in the mmapv1 pre hook method.

hooks/pre/30-upgrade.rb
```diff
@@ -80,6 +80,32 @@ def fix_katello_settings_file
...
+# rubocop:disable MethodLength
+def mongo_mmapv1_check
+  custom_hiera = '/etc/foreman-installer/custom-hiera.yaml'
+  mongodb_dir = '/var/lib/mongodb/'
+  # check and see if we have a pulp_database already.
+  if File.file?("#{mongodb_dir}/pulp_database.0")
+    # check if we have modifed the custom_hiera file or if there is a wiredTiger file in the db directory.
+    if File.foreach("#{custom_hiera}").grep(/mongodb::server::storage_engine:/).any? || File.file?("#{mongodb_dir}/WiredTiger.wt")
+      logger.info 'No changed needed, Mongo storage engine will installed/kept with wiredTiger'
+    else
```
ehelms: I think we may want to stop mongod in this else block.

hooks/post/30-upgrade.rb
```diff
@@ -137,6 +138,20 @@ def remove_event_queue
   end
 end
 
+def remove_legacy_mongo
+  # Check to see if the RPMS exist and if so remove them and create the upgrade done file.
+  if `rpm -q mongodb --queryformat=%{version}`.start_with?('2.')
+    logger.warn 'removing Mongo 2.x packages, config and log files.'
+    Kafo::Helpers.execute("yum remove -y mongodb-2* mongodb-server-2*, false")
```
ehelms: I still have mongodb-server installed:

```
mongodb-server.x86_64
```

hooks/pre/30-upgrade.rb
```diff
@@ -212,6 +240,7 @@ def fail_and_exit(message)
   if katello || foreman_proxy_content
     upgrade_step :upgrade_qpid_paths
     upgrade_step :migrate_pulp, :run_always => true
+    upgrade_step :mongo_mmapv1_check
```
ehelms: From my testing, this needs to be the last step we take, I moved it to line 258.

PR #611, opened 2018-04-17
Fixes #23292 - make s390x ks repos bootable

sean797:


sean797:
Thanks, rebased

sean797:
@jlsherrill that wasn't the case on the most recent discussion. 
https://community.theforeman.org/t/database-and-service-actions-in-rpm-post-scripts/7311/6

My understanding of the outcome:
- Some users want migrations run in `%postrans`
- Some users DO NOT want migrations run in `%postrans` (Foreman clusters would be one use-case)
- Lets extracts it out to another script (maybe foreman-maintain?) and `%postrans` can call the script
- The script can actaully run migrations (or not) based on user configuration

ekohl:
@jlsherrill my thoughts about core was a `foreman-rake upgrade` which plugins could extend. It would be similar to the model we have here in the installer and plugins could add tasks into it. Idempotency is an important one. Then we can call it either in `%posttrans` (possibly guarded with a setting), `foreman-installer`, `foreman-maintain` or a containerization tool.

jlsherrill:
ah i see, i must have been thinking about a past conversation.  I think that plan sounds good.  I'm hoping that our team can tackle this work very soon

ehelms:
@sean797 Mind doing a rebase?

sean797:
rebased :+1:

ekohl:
Thanks!

PR #612, opened 2018-04-18
Fixes #23242 - Add cli plugins to Katello scenario

chris1984:


chris1984:
Sounds good, I will add them and resubmit the changes.

chris1984:
@stbenjam updated.

ekohl:
@stbenjam I indeed was starting to add them all. https://github.com/theforeman/puppet-foreman/commit/b04778a10cc0b7bb83e8e3d3cdc9e7ae857d2126 should make it easy to add all plugins. I was starting to look into a way to automatically install matching CLI plugins for the normal plugins but haven't found one (though I didn't spend much time on it).

PR #613, opened 2018-04-20
Fixes #23345 - only call cleanup if we could initialize

evgeni:
otherwise the following error is raised

    $ foreman-proxy-certs-generate --help
    /usr/sbin/foreman-proxy-certs-generate:88:in `ensure in <main>': undefined method `cleanup' for nil:NilClass (NoMethodError)
	    from /usr/sbin/foreman-proxy-certs-generate:88:in `<main>'

and that masks the real error that happened in initialize


chris1984:
Thanks @evgeni

PR #617, opened 2018-05-01
Fixes #23470 - Reset puppet server version to ensure smooth upgrades

ehelms:


ehelms:
@ekohl This might seem odd at first, but this is needed for example if you have Puppet 4 and move to the Puppet 5 repository while upgrading which I think is a supported path for us now that P5 is not drastic from P4.

hooks/pre_validations/31-upgrade-puppet.rb
```diff
@@ -1,23 +1,7 @@
...
-  katello = Kafo::Helpers.module_enabled?(@kafo, 'katello')
-  foreman_proxy_content = Kafo::Helpers.module_enabled?(@kafo, 'foreman_proxy_content')
-
-  fail_and_exit 'Puppet 4 to 5 upgrade is not currently supported for the chosen scenario.' unless katello || foreman_proxy_content
-
-  Kafo::Helpers.log_and_say :info, 'Resetting puppet params...'
+if app_value(:upgrade)
+  Kafo::Helpers.log_and_say :info, 'Resetting puppet server version param...'
 
   if !app_value(:noop)
```
ekohl: Maybe also make this `unless` rather than `if !`

hooks/boot/31-upgrade-puppet.rb
```diff
@@ -1,6 +0,0 @@
-app_option(
-  '--upgrade-puppet',
-  :flag,
-  "Run the steps neccessary to upgrade from Puppet 3 to Puppet 4.",
```
ekohl: Looks like we never updated this :)

ekohl:
> @ekohl This might seem odd at first, but this is needed for example if you have Puppet 4 and move to the Puppet 5 repository while upgrading which I think is a supported path for us now that P5 is not drastic from P4.

As I mentioned: reinstalling the oauth gem is needed from puppet 4 to puppet 5.

ehelms:
Updated to run this at all times and check the environment. Tested on an upgrade and worked.

PR #619, opened 2018-05-04
Refs #23470 - Ensure puppetserver_version param exists

ehelms:


hooks/pre_validations/31-upgrade-puppet.rb
```diff
@@ -7,5 +7,7 @@
     end
   end
 
-  Kafo::Helpers.reset_value(param('puppet', 'server_puppetserver_version'))
+  if puppetserver_version_param = param('puppet', 'server_puppetserver_version')
+    Kafo::Helpers.reset_value(puppetserver_version_param)
```
ekohl: I think `puppetserver_version_param.unset_value` is good enough. We should remove `reset_value`

PR #623, opened 2018-05-07
Fixes #23526 - remove elasticsearch_message upgrade step

evgeni:


evgeni:
@ekohl you mean this *should* be added to the release notes, or it is in the rel notes and is sufficient there?

ekohl:
> @ekohl you mean this should be added to the release notes, or it is in the rel notes and is sufficient there?

I don't think it's in the release notes now but it should be added that we no longer remove elastic search.

evgeni:
Uh, where are release notes stored? :)

ekohl:
https://github.com/theforeman/theforeman.org/blob/gh-pages/plugins/katello/nightly/release_notes/release_notes.md

evgeni:
https://github.com/theforeman/theforeman.org/pull/1076

PR #624, opened 2018-05-10
fixes #23375 - use db names from puppet params

stbenjam:
To test, install a katello like this:

```
foreman-installer --scenario katello --katello-candlepin-db-name platypus --katello-pulp-db-name capybara
```

Run foreman-installer --reset with this PR, and notice it drops the correct databases.


chris1984:
Thanks @stbenjam

stbenjam:
Thanks!

PR #628, opened 2018-05-15
Fixes #23622 - allow generating certs_tar for Katello app

sean797:
Requires https://github.com/theforeman/puppet-certs/pull/199

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,8 +42,9 @@ foreman-installer --scenario                     "katello"\
                   --foreman-proxy-puppetca       "false"
 
 # Generate certificates for installing Foreman proxy on another system
-foreman-proxy-certs-generate --foreman-proxy-fqdn "myforeman-proxy-content.example.com"\
-                                     --certs-tar    ~/myforeman-proxy-content.example.com-certs.tar
+katello-generate-cert-archive --server-fqdn "myforeman-proxy-content.example.com"\
+                              --certs-tar    "~/myforeman-proxy-content.example.com-certs.tar"
+                              --foreman-proxy true
 
 # Copy the ~/myforeman-proxy.example.com-certs.tar to the foreman-proxy system
 # register the system to Katello and run:
@@ -67,8 +68,12 @@ foreman-installer --scenario                                "foreman-proxy-conte
                   --foreman-proxy-dhcp                      "true"\
                   --foreman-proxy-dhcp-interface            "virbr1"\
                   --foreman-proxy-tftp                      "true"
-```
 
+# Generate certificates for installing Foreman Application on another system
+katello-generate-cert-archive --server-fqdn "myforeman1.example.com"\
+                              --certs-tar    "~/myforeman1.example.com-certs.tar"
+                              --foreman-application true
+```
 ## Data Reset
 
 If you run into an error during the initial installation or wish to reset your installation entirely, the installer provides a reset option. To reset the database and all subsystems:
--- a/Rakefile
+++ b/Rakefile
@@ -54,6 +54,7 @@ end
 task :generate_parser_caches => [PARSER_CACHE_DIR] do
   caches = [
     "#{PARSER_CACHE_DIR}/katello.yaml",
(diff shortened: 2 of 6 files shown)


Rakefile
```diff
@@ -54,18 +54,23 @@ end
 task :generate_parser_caches => [PARSER_CACHE_DIR] do
   caches = [
     "#{PARSER_CACHE_DIR}/katello.yaml",
+    "#{PARSER_CACHE_DIR}/cert_generate_archive.yaml"
```
beav: needs a comma here

sean797:
Thanks @beav updated! Tests won't pass because https://github.com/theforeman/puppet-certs/pull/199 is required anyway :)

README.md
```diff
@@ -42,8 +42,9 @@ foreman-installer --scenario                     "katello"\
                   --foreman-proxy-puppetca       "false"
 
 # Generate certificates for installing Foreman proxy on another system
-foreman-proxy-certs-generate --foreman-proxy-fqdn "myforeman-proxy-content.example.com"\
-                                     --certs-tar    ~/myforeman-proxy-content.example.com-certs.tar
+katello-generate-cert-archive --server-fqdn "myforeman-proxy-content.example.com"\
```
ehelms: Could we drop the katello naming aspect?

hooks/boot/02-message-helpers.rb
```diff
@@ -67,18 +82,68 @@ def proxy_instructions_message(kafo)
...
      parameters, see <%= color("foreman-installer --scenario foreman-proxy-content --help", :info) %> and
      documentation for more info on setting up additional services):
 
+MSG
+    end
+
+    def foreman_app_instructions_message(certs_tar, server_fqdn)
+      candlepin_oauth_key = Kafo::Helpers.read_cache_data("candlepin_oauth_secret")
+      candlepin_host = `hostname -f`.strip
+      say <<MSG
```
ehelms: Empty line above here would help with code readability

ehelms:
Is it fair to say the current implementation here aims to move away from the use of foreman-proxy-certs-generate and introduce a new tool but maintain both?

sean797:
> Is it fair to say the current implementation here aims to move away from the use of foreman-proxy-certs-generate and introduce a new tool but maintain both?

yes - assuming everyone is happy with that?

README.md
sean797: So just ‘generate-cert-archive’ ?

I think the katello name is important as it’s explicit for what it is for. It’s also consistent with other tools (backup, restore, service, etc... )

ehelms: How about `foreman-generate-cert-archive` ?

ehelms:
Any reason not to simply do a full rename here and now? This is typically a one off script that is not part of automation. Thus the risk to a full rename to me seems minimal compared to carrying technical debt.

sean797:
I'm happy to just rename if we are sure? I use it in [automation](https://github.com/sean797/ansible-role-foreman_installer/blob/master/tasks/katello-proxy-certs.yml#L94) and its used in forklift, so I doubt we are the only ones doing automation with it, hence the deprecation

README.md
sean797: Does the Foreman name make sense if the certs archive cannot be used with vanilla foreman? If there is a plan to make it compatiable with Foreman then, I'd be happier naming it that. 

I think the current naming can make the two "brands" (Katello & Foreman) confusing to users, sometimes we swap the names around for no good reason. To me at least the Foreman name implies I would use it with Foreman, but in this case the tool cannot be used with vanilla Foreman.

zjhuntin:
@sean797 @ehelms is there anymore discussion to this PR? I see https://github.com/theforeman/puppet-certs/pull/210 as well, just wondering as it's related to a 3.9 redmine.

sean797:
@zjhuntin since you updated the issue I assume you had a conversation away from here? I'd really like to see the work for split deployments start gaining some traction soon. IMO this issue (the certs aspect) is ready.

ekohl:

PR #632, opened 2018-05-15
Fixes #22598 - add check for key usage Key Encipherment

jaredjennings:
I obtained a new server certificate for my Satellite server, but then found that yum couldn't fetch repository updates (errno 14, curl#60, NSS -8102, SEC_ERROR_INADEQUATE_KEY_USAGE). The certificate had X509v3 Key Usages of Digital Signature and Non Repudiation. When I got a new certificate which also included the Key Encipherment key usage, yum worked. (Red Hat Support case 02099325.)

This PR adds a check to katello-certs-check to ensure that the server certificate has the Key Encipherment key usage. With this patch, my old certificate fails katello-certs-check, and my new certificate passes.

### DIFF
--- a/bin/katello-certs-check
+++ b/bin/katello-certs-check
@@ -145,6 +145,16 @@ function check-cert-san () {
     fi
 }
 
+function check-cert-usage-key-encipherment () {
+    printf "Checking Key Usage extension on certificate for Key Encipherment"
+    CHECK=$(openssl x509 -noout -text -in $CERT_FILE | grep -A1 'X509v3 Key Usage:' | grep 'Key Encipherment')
+    if [ $? == "0" ]; then
+        success
+    else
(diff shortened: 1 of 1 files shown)


jaredjennings:
Do I need to file a Foreman ticket?

chris1984:
@jaredjennings if you can and modify the commit/pr message to be like

~~~
Fixes #ticketnumber - title
~~~

jaredjennings:
@chris1984 under which project should I file the bug? Katello? foreman-installer?

chris1984:
@jaredjennings you want to do Katello then category installer.

jaredjennings:
Ah never mind, found some existing katello-certs-check issues filed against "Installer" category.

chris1984:
Thanks for updating it, still trying to create a cert so i can reproduce this

chris1984:
Thanks @jaredjennings

PR #643, opened 2018-06-01
Fixes #23766 - remove csr references from certs-check

chris1984:


iNecas:
Seems like there was some missed remove of the `-r` option in https://github.com/Katello/katello-installer/blob/2e213ef74362c5f0c8aa0a60d181ef1c776d7df0/hooks/pre/20-certs_update.rb#L59, see https://bugzilla.redhat.com/show_bug.cgi?id=1616228

ekohl:
https://github.com/Katello/katello-installer/pull/671

PR #648, opened 2018-06-15
Refs #23030 - Update logic on Mongo pre check

chris1984:


hooks/pre/30-upgrade.rb
```diff
@@ -69,6 +69,8 @@ def mongo_mmapv1_check
         file.write("Mongo storage engine set to mmapv1 on #{Time.now}")
       end
     end
+  else
+    logger.info 'No changed needed, Mongo storage engine will installed/kept with WiredTiger.'
```
ekohl: IMHO this is at most a debug.

chris1984:
@ekohl updated.

PR #649, opened 2018-06-18
Bump puppet-mongodb to 2.2.0

ekohl:
This includes a fix for the is_master fact to gracefully handle a non-existing config file.


ekohl:
Looks like the tag wasn't actually pushed yet. Automation broke and now someone with the right permissions needs to do it manually. Will repush when that's done.

ekohl:
The tag has been pushed and is now published on the forge.

chris1984:
ACK

PR #652, opened 2018-06-27
Fixes #24088 - Maintain the puppet answer in foreman_proxy_content

ekohl:
This answer is still needed to determine whether the certs for puppet integration need to be placed.


ehelms:
Changing the migrations seems bad? and won't necessarily solve any problems since this was added many releases ago?

ekohl:
That's a fair point. Probably better to detect the current answer for Puppet and copy that in a new migration.

ekohl:
Updated as a new migration now that recalculates the value.

ehelms:
@stbenjam Mind reviewing this? You've always been better at spotting migration edge cases than me.

PR #656, opened 2018-07-02
Refs #24066 - Update Rubocop method length cop

chris1984:
Allow longer method length for MongoDB Engine Upgrade


chris1984:
Had to update foreman-proxy-certs-generate to keep Travis happy.

PR #657, opened 2018-07-06
Fixes #24189 - switch to foreman upgrade task

jlsherrill:


hooks/post/30-upgrade.rb
```diff
@@ -171,30 +80,9 @@ def fail_and_exit(message)
...
-      upgrade_step :set_virt_who_on_pools, :long_running => true
-      upgrade_step :remove_unused_products, :long_running => true
-      upgrade_step :create_host_subscription_associations, :long_running => true
-      upgrade_step :reindex_docker_tags, :long_running => true
-      upgrade_step :republish_file_repos, :long_running => true
-      upgrade_step :import_backend_consumer_attributes, :long_running => true
-      upgrade_step :remove_registration_tasks
       upgrade_step :remove_legacy_mongo
-      upgrade_step :ensure_ks_repos_are_bootable, :long_running => true
+      upgrade_tasks
```
jlsherrill: @ekohl I wasn't sure if this is how you expected this to be run, or if we could re-use a hook in foreman's installer (and i guess add a --upgrade flag to it?).

ekohl: Since this is in Foreman core I think this should be done if the Foreman module is enabled rather than just katello. For database migrations we also have a setting that indicates a migration is needed. The same goes for seeds. See https://github.com/theforeman/puppet-foreman/blob/b937165001aac788fbafef0422984a59edaeba87/manifests/database.pp#L33-L46. I think this would be good thing for the upgrade as well. It could even be done in pure Puppet similar to `db:migrate` and `db:seed` (yes, in katello-installer we run migration and seeding twice now) but it'd be fine to do that in a phase 2.

jlsherrill: @ekohl  for this 'first phase' should this live in foreman-installer?  Or are you just saying to leave it here and check for the foreman module.

ekohl: It's fine to let it live here initially.

jlsherrill:
@ekohl this is now ready~

hooks/post/30-upgrade.rb
```diff
@@ -180,30 +85,8 @@ def fail_and_exit(message)
...
-      upgrade_step :import_rpms, :long_running => true
-      upgrade_step :import_distributions, :long_running => true
-      upgrade_step :import_puppet_modules, :long_running => true
-      upgrade_step :import_subscriptions, :long_running => true
-      upgrade_step :import_product_content, :long_running => true
-      upgrade_step :add_export_distributor, :long_running => true
-      upgrade_step :remove_docker_v1_content, :long_running => true
-      upgrade_step :update_puppet_repository_distributors, :long_running => true
-      upgrade_step :update_subscription_facet_backend_data, :long_running => true
       upgrade_step :remove_event_queue
```
ekohl: Why do you leave this in?

jlsherrill: @ekohl I don't think there were any upgrade rake tasks added for 3.8, so we technically will only require them to be at 3.7 at a minimum, which seems fairly safe to do now.    Most of these were dropped, but all the :always_run ones were moved to katello here: https://github.com/Katello/katello/pull/7504

I left the remove_event_queue because it is not a rake task (and not really related to any of my work).  I don't mind removing it if its not needed anymore though.

ekohl: I'd be in favor of removing it as well. IMHO `db_seed` should also be removed. The installer should take care of this and if it doesn't that's a bug. Given the new rake task will hard fail on this, we'll be automatically notified by this in our pipelines.

jlsherrill: done!

akofink:
FYI, I'm adding a 3.8(.1) upgrade task in https://github.com/Katello/katello/pull/7655.

hooks/post/30-upgrade.rb
```diff
@@ -178,32 +74,7 @@ def fail_and_exit(message)
     upgrade_step :restart_services, :run_always => true
 
     if Kafo::Helpers.module_enabled?(@kafo, 'katello')
```
ekohl: This this be generalized to just `module_enabled?('foreman')`? It should use the [kafo hooks DSL](https://github.com/theforeman/kafo#hooks) and since the rake task is in core foreman it means we can later easily port it to Foreman.

jlsherrill: works4me!

hooks/post/30-upgrade.rb
```diff
@@ -177,33 +73,8 @@ def fail_and_exit(message)
...
-      upgrade_step :republish_file_repos, :long_running => true
-      upgrade_step :import_backend_consumer_attributes, :long_running => true
-      upgrade_step :remove_registration_tasks
-      upgrade_step :ensure_ks_repos_are_bootable, :long_running => true
-      upgrade_step :set_upstream_pool_id, :long_running => true
+    if module_enabled?('foreman')
+      upgrade_tasks
     end
 
     if [0, 2].include? @kafo.exit_code
```
ekohl: Looking at the now greatly simplified code: is this redundant? I think `exit` already stops the process so this would be always true I think.

jlsherrill: :+1:

hooks/post/30-upgrade.rb
```diff
@@ -11,113 +11,9 @@ def db_seed
   Kafo::Helpers.execute('foreman-rake db:seed')
```
ekohl: This function is now no longer needed

hooks/post/30-upgrade.rb
```diff
@@ -176,39 +72,7 @@ def fail_and_exit(message)
 if app_value(:upgrade)
   if [0, 2].include?(@kafo.exit_code)
     upgrade_step :restart_services, :run_always => true
```
ekohl: IIRC we stop all services before upgrading and the installer should start all services it needs (or restart if anything changed). That makes me wonder if this is really needed.

jlsherrill: @ehelms your thoughts on this?

ehelms: I tend to agree we should be able to drop this

jlsherrill: any idea why we check for a return code of 2?

jlsherrill:
@ekohl done!

hooks/post/30-upgrade.rb
```diff
@@ -175,40 +63,7 @@ def fail_and_exit(message)
...
-      upgrade_step :reindex_docker_tags, :long_running => true
-      upgrade_step :republish_file_repos, :long_running => true
-      upgrade_step :import_backend_consumer_attributes, :long_running => true
-      upgrade_step :remove_registration_tasks
-      upgrade_step :ensure_ks_repos_are_bootable, :long_running => true
-      upgrade_step :set_upstream_pool_id, :long_running => true
-    end
-
-    if [0, 2].include? @kafo.exit_code
-      Kafo::Helpers.log_and_say :info, 'Upgrade completed!'
```
ekohl: Sorry if I was unclear, but I meant that the `if` was redundant. The `Upgrade completed!` can still be useful though I don't mind a silent installer :)

jlsherrill: added it back below :)

PR #661, opened 2018-07-17
Fixes #24272 - reimport subscriptions when upgrading to 3.7

jlsherrill:


jlsherrill:
An additional column was added here: https://github.com/Katello/katello/commit/b3efb0ebda8d56628d2a92cfe2f14e24e63bc1c4#diff-81ae82155cdf6fe8e45be63d30434815

but this needs to be reimported on upgrade in order to populate it.

jlsherrill:
hopefully the last one of these we need to do :)

beav:
ACK, thanks @jlsherrill !

I asked on IRC about why we are calling `foreman-rake katello:import_subscriptions` again if it's already happening. The issue is that if a task is defined here, it only happens once and does not run again on subsequent upgrades unless `always_run` is set to true. Thus, if you add a DB migration that requires import_subs to run again to populate a new field, you'll need to add another invocation of import_subs here.

This is being reworked, but for right now redefining is the way to go.

PR #662, opened 2018-07-17
Pin installer modules

ehelms:


ehelms:
@ekohl please to verify

Puppetfile
```diff
@@ -1,12 +1,9 @@
...
-mod 'katello/candlepin',             :git => 'https://github.com/theforeman/puppet-candlepin'
-mod 'katello/foreman_proxy_content', :git => 'https://github.com/theforeman/puppet-foreman_proxy_content'
-mod 'katello/certs',                 :git => 'https://github.com/theforeman/puppet-certs'
-mod 'katello/katello',               :git => 'https://github.com/theforeman/puppet-katello'
-mod 'katello/pulp',                  :git => 'https://github.com/theforeman/puppet-pulp'
-mod 'katello/qpid',                  :git => 'https://github.com/theforeman/puppet-qpid'
-
-# A fork that marks metadata as compatible with puppet 5
-# https://github.com/thias/puppet-squid3/pull/59
-mod 'thias-squid3',                  :git => 'https://github.com/theforeman/puppet-squid3', :ref => 'allow-puppet-5'
```
ekohl: This one is still needed because we use a fork. I've started to work on replacing this module but that's still WIP.

Puppetfile
```diff
@@ -1,12 +1,9 @@
...
+mod 'katello/candlepin', '>= 5.1.1 < 5.2.0'
```
ekohl: In Foreman I've kept the alignment. Perhaps do it here too? I've recently thought of how I could actually align it automatically but till that time ... :)

Rakefile
```diff
@@ -21,7 +21,7 @@ end
 BUILDDIR = File.expand_path(ENV['BUILDDIR'] || '_build')
 PKGDIR = ENV['PKGDIR'] || File.expand_path('pkg')
 FOREMAN_MODULES_DIR = File.expand_path(ENV['FOREMAN_MODULES_DIR'] || '/usr/share/foreman-installer/modules')
-FOREMAN_BRANCH = ENV['FOREMAN_BRANCH'] || 'develop'
+FOREMAN_BRANCH = '1.19-stable'
```
ekohl: Maybe keep the `ENV` option but just replace the default?

ehelms: I don't see value in the ENV option to be honest. Let's lock it down.

Puppetfile
ehelms: Heh, I just used the `pin_modules` task so I guess I can go edit the output after :)

Puppetfile
ehelms: Heh, I just used the `pin_modules` task. It may need to be smarter.

Rakefile
ehelms: I suppose some other distro of this could use it. Its put back.

ehelms:
@ekohl Thanks for the review. Updated.

Puppetfile
ekohl: Perhaps if there's a ref then it should keep it.

ehelms:
@ekohl fixed

Puppetfile
```diff
@@ -1,12 +1,9 @@
...
-# A fork that marks metadata as compatible with puppet 5
-# https://github.com/thias/puppet-squid3/pull/59
-mod 'thias-squid3',                  :git => 'https://github.com/theforeman/puppet-squid3', :ref => 'allow-puppet-5'
+mod 'katello/candlepin',              '>= 5.1.1 < 5.2.0'
+mod 'katello/foreman_proxy_content',  '>= 7.2.0 < 7.3.0'
+mod 'katello/certs',                  '>= 4.2.0 < 4.3.0'
+mod 'katello/katello',                '>= 8.0.1 < 8.1.0'
+mod 'katello/pulp',                   '>= 5.5.0 < 5.6.0'
+mod 'katello/qpid',                   '>= 4.2.2 < 4.3.0'
+mod 'thias-squid3',                   :git => 'https://github.com/theforeman/puppet-squid3', :ref => 'allow-puppet-5'
```
ekohl: I'd appreciate keeping the comment for this odd one in. While you're at it, could you decreasing indenting by 1 so it matches the previous indenting?

PR #671, opened 2018-08-16
Fixes #24632 - cert-req needs to be removed

ahumbe:
Now katello-certs-check command does not accept -r ( request file) so this option should be removed from the 20-certs_update.rb hook.


ekohl:
Thanks!

PR #673, opened 2018-08-16
prepare 3.7.1 release

evgeni:
this is my attempt to help you with the 3.7.1 release, shout at me if I did smth wrong :)

I wanted to have the following fixes in:
* https://projects.theforeman.org/issues/24574 (does not have "target version" set, but I'd like to have it in ;))
* https://github.com/theforeman/puppet-qpid/pull/96

the `Puppetfile.lock` diff is longer than I've expected it to be, but I have no idea why


ekohl:
https://github.com/Katello/katello-installer/pull/671 is probably also a good candidate for 3.7.1.

Puppetfile.lock
```diff
@@ -129,21 +113,12 @@ GIT
...
-  puppetlabs-mysql (>= 4.0.0)
-  puppetlabs-postgresql (>= 5.6.0)
-  puppetlabs-puppetdb (>= 0)
-  theforeman-dhcp (< 4.2.0, >= 4.1.1)
-  theforeman-dns (< 5.3.0, >= 5.2.0)
-  theforeman-foreman (< 9.3.0, >= 9.2.0)
-  theforeman-foreman_proxy (< 7.3.0, >= 7.2.1)
-  theforeman-git (< 4.1.0, >= 4.0.0)
-  theforeman-puppet (< 9.1.0, >= 9.0.0)
-  theforeman-tftp (< 4.2.0, >= 4.1.0)
```
ehelms: I think you want to put these back

evgeni: jaja, I thought librarian would be smarter than this ;)

evgeni:
I'd question the usage (and documentation) of librarian, if it's not useful in our case... /me goes modifying files by hand.

PR #639, opened 2018-05-29
Fixes #24066 - Add Mongo storage engine upgrade

chris1984:

PR #674, opened 2018-08-20
Refs #24066 - Update Rubocop method length cop.

chris1984:


chris1984:
Updating cop length to make tests green on https://github.com/Katello/katello-installer/pull/639

ekohl:
I don't think we should keep increasing the max length globally. If we blindly increase it every time we write a longer method we might as well disable the whole cop.

I'd recommend to add local overrides for this method or introduce helpers for parts of the function.

chris1984:
Sounds good, ill close it out and do that on the other pr.

PR #677, opened 2018-08-30
Fixes #24752 - reset server_puppetserver_metrics

evgeni:


evgeni:
What I don't totally grasp: why did it only start failing recently? https://ci.centos.org/job/foreman-katello-upgrade-nightly-test/27/ was still fine. And it has the same puppetserver RPM.

PR #679, opened 2018-08-30
Fixes #24752 - reset server_puppetserver_metrics

evgeni:
(cherry picked from commit 08d694639c2b8a752fa6c8166dea10d5a6fed2c9)


evgeni:
Or should I use `Kafo::Helpers.reset_value` for 3.7?

ekohl:
Both should work. I'd like to get rid of reset value so I'm fine with this.

PR #683, opened 2018-09-11
No need to remove MongoDB log files

sean797:
ping @chris1984

### DIFF
--- a/hooks/post/30-upgrade.rb
+++ b/hooks/post/30-upgrade.rb
@@ -15,7 +15,7 @@ def remove_legacy_mongo
   if `rpm -q mongodb --queryformat=%{version}`.start_with?('2.')
     logger.warn 'removing MongoDB 2.x packages, config and log files.'
     Kafo::Helpers.execute('yum remove -y mongodb-2* mongodb-server-2* > /dev/null 2>&1')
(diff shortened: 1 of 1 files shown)


chris1984:
@sean797 Thanks for catching that.

hooks/post/30-upgrade.rb
```diff
@@ -15,7 +15,7 @@ def remove_legacy_mongo
   if `rpm -q mongodb --queryformat=%{version}`.start_with?('2.')
     logger.warn 'removing MongoDB 2.x packages, config and log files.'
     Kafo::Helpers.execute('yum remove -y mongodb-2* mongodb-server-2* > /dev/null 2>&1')
-    Kafo::Helpers.execute('rm -rf /etc/mongod.conf /var/log/mongodb')
+    Kafo::Helpers.execute('rm -rf /etc/mongod.conf')
```
ekohl: Looking at this, why do we use `-r` on a single file? That's just asking for errors. We could also use pure ruby with `File.unlink('/etc/mongod.conf') if File.exist?('/etc/mongod.conf')`

chris1984: Good catch, sorry for merging so fast.

chris1984: I will make a new PR to fix it

PR #686, opened 2018-09-18
Fixes #24967 - Move mongo upgrade inside upgrade if logic.

chris1984:


chris1984:
@ekohl are these ok to merge?

chris1984:
Not Needed.

PR #688, opened 2018-09-23
Fixes #25011 - only start PostgreSQL in the Katello scenario

evgeni:
There is no PostgreSQL required on the Proxy.


mccun934:
why is this effecting us now vs in older versions?

evgeni:
@mccun934 we switched to foreman-maintain instead of katello-service.

f-m errors out when no services match the request (https://github.com/theforeman/foreman_maintain/blob/master/definitions/features/service.rb#L30), which i think is the correct behavior.

k-s was more relaxed here.

chris1984:
@evgeni Thanks!

PR #691, opened 2018-09-27
Fixes #25056 - Remove mappings

ekohl:
Kafo hasn't needed these trivial mappings and we eventually clear them
anyway.

### DIFF
--- a/config/foreman-proxy-content.migrations/160120212700-foreman-proxy.rb
+++ b/config/foreman-proxy-content.migrations/160120212700-foreman-proxy.rb
@@ -1,13 +1,5 @@
 # Decouple puppet-foreman_proxy from puppet-capsule
 
-def mod(manifest, params)
-  {
-    :manifest_name => manifest,
-    :params_name => params,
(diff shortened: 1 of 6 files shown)


ehelms:
When were mappings removed? I tend not to like back editing migrations unless we are absolutely sure within supported versions this won't break anything.

ekohl:
For katello we clear it here: https://github.com/Katello/katello-installer/blob/d9e5abbb2a5825cac582e3518364718f9c6549bf/config/katello.migrations/161124183722-remove-mappings.rb#L1

For the proxy it's here: https://github.com/Katello/katello-installer/blob/d9e5abbb2a5825cac582e3518364718f9c6549bf/config/foreman-proxy-content.migrations/161124183740-remove-mappings.rb#L1

As you can see by the filenames these are executed after the migrations we modify meaning they are fully redundant. e27ac7386a74150f3ceaaea26a07f1e98de54a8a originally introduced those and it's dated in 2016.

PR #694, opened 2018-10-01
Refs #24966 - Move mongo syspath to after upgrade

chris1984:


hooks/post/31-mongo_syspath.rb
```diff
@@ -0,0 +1,11 @@
...
+  `rpm -qa | grep rh-mongodb34-syspaths`
+  if $?.success?
+    logger.info 'rh-mongodb34-syspaths already installed, skipping.'
+  else
+    logger.info 'rh-mongodb34-syspaths not present, installing.'
+    Kafo::Helpers.execute('yum install -y -q rh-mongodb34-syspaths')
+  end
+end
+
+install_syspaths
```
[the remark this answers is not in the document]
chris1984: Updated

hooks/post/31-mongo_syspath.rb
```diff
@@ -0,0 +1,15 @@
...
+    logger.info 'rh-mongodb34-syspaths not present, installing.'
+    Kafo::Helpers.execute('yum install -y -q rh-mongodb34-syspaths')
+  end
+end
+
+# Fail if we are not using the Katello/Foreman Proxy Content scenario.
+katello = Kafo::Helpers.module_enabled?(@kafo, 'katello')
+foreman_proxy_content = Kafo::Helpers.module_enabled?(@kafo, 'foreman_proxy_content')
+logger.error 'MongoDB storage engine upgrade is not currently supported for the chosen scenario.' unless katello || foreman_proxy_content
+install_syspaths
```
ehelms: I think this line needs to be inside the if check on module enabled as well. Otherwise it will log the above message and still call this code.

chris1984:
@ehelms updated.

hooks/post/31-mongo_syspath.rb
```diff
@@ -0,0 +1,23 @@
...
+  end
+end
+
+# Fail if we are not using the Katello/Foreman Proxy Content scenario.
+katello = Kafo::Helpers.module_enabled?(@kafo, 'katello')
+foreman_proxy_content = Kafo::Helpers.module_enabled?(@kafo, 'foreman_proxy_content')
+if katello || foreman_proxy_content
+  install_syspaths
+else
+  fail_and_exit 'MongoDB storage engine upgrade is not currently supported for the chosen scenario.'
```
ehelms: Should we fail here or just warn that a scenario that is not supported was detected and thus mongodb-syspaths was not installed? The likely hood of hitting this hook under that condition is slim but I think thats cleaner than hard fail.

chris1984: +1 to warn

ekohl: IMHO it should be quiet. The certificate generation script also calls hooks so I think that failing here breaks that functionality.

hooks/post/31-mongo_syspath.rb
```diff
@@ -0,0 +1,18 @@
...
+  end
+end
+
+# Fail if we are not using the Katello/Foreman Proxy Content scenario.
+katello = Kafo::Helpers.module_enabled?(@kafo, 'katello')
+foreman_proxy_content = Kafo::Helpers.module_enabled?(@kafo, 'foreman_proxy_content')
+if katello || foreman_proxy_content
+  install_syspaths
+else
+  logger.warn 'Selected scenario is not correct for rh-mongodb34-syspaths install, skipping'
```
ekohl: Is this really a warning? Remember that `foreman-proxy-certs-generate` also executes this code. That means you will fill logs with warnings that say nothing. IMHO this is at most a debug.

hooks/post/31-mongo_syspath.rb
```diff
@@ -0,0 +1,18 @@
...
+  `rpm -qa | grep rh-mongodb34-syspaths`
+  if $?.success?
+    logger.debug 'rh-mongodb34-syspaths already installed, skipping.'
+  else
+    logger.info 'rh-mongodb34-syspaths not present, installing.'
+    Kafo::Helpers.execute('yum install -y -q rh-mongodb34-syspaths')
+  end
+end
+
+# Fail if we are not using the Katello/Foreman Proxy Content scenario.
```
ekohl: Personally I think the code is self-documenting and could see this removed but it should at least match the code

hooks/post/31-mongo_syspath.rb
```diff
@@ -0,0 +1,17 @@
...
+    Kafo::Helpers.execute('yum install -y -q rh-mongodb34-syspaths')
+  end
+end
+
+katello = Kafo::Helpers.module_enabled?(@kafo, 'katello')
+foreman_proxy_content = Kafo::Helpers.module_enabled?(@kafo, 'foreman_proxy_content')
+if katello || foreman_proxy_content
+  install_syspaths
+else
+  logger.debug 'Selected scenario is not correct for rh-mongodb34-syspaths install, skipping'
```
ekohl: `not correct` is IMHO incorrect because a user could read and think something is wrong. I'd say `not applicable`.

chris1984:
@ekohl updated.

PR #705, opened 2018-10-24
set prerelease in version string

komidore64:


VERSION
```diff
@@ -1 +1 @@
-3.10.0
+3.10.0.master
```
ehelms: I guess we have to ask ourselves if we want to follow existing methods and use a dash -- https://github.com/theforeman/foreman-installer/blob/develop/VERSION#L1

komidore64: I'm fine to stick with convention. Apologies, I just looked at what [katello-installer-base.spec](https://github.com/theforeman/foreman-packaging/blob/rpm/develop/packages/katello/katello-installer-base/katello-installer-base.spec#L15) looked for.

komidore64:
@ehelms updated

chris1984:
Thanks @komidore64

PR #707, opened 2018-10-29
Fixes #25336 - Move remove MongoDB from post to pre hook.

chris1984:


chris1984:
Tested it already, removing Needs-Testing Label

hooks/pre/30-upgrade.rb
```diff
@@ -73,6 +74,23 @@ def mongo_mmapv1_check
   end
 end
 
+def remove_legacy_mongo
+  # Check to see if the RPMS exist and if so remove them and create the upgrade done file, and install rh-mongodb34-mongodb-syspaths.
+  return if File.exist?(MONGO_REMOVAL_COMPLETE)
```
ekohl: I'm fine with keeping this in since it was there before, but I do wonder if it's really needed. Using `upgrade_step` means it's already ran only once.

chris1984: +1 for taking it out

PR #711, opened 2018-11-02
Fixes #25377 - Update capsule_tar check to include ssl-build

chris1984:


chris1984:
[test prprocessor]

hooks/pre_validations/12-check_capsule_tar.rb
```diff
@@ -4,12 +4,14 @@ def error(message)
...
 
+ssl_directory = param('certs', 'ssl_build_dir')
+
 certs_tar = param('foreman_proxy_certs', 'certs_tar') ||
   param('foreman_proxy_content', 'certs_tar')
 
 if certs_tar.value
   certs_tar.value = File.expand_path(certs_tar.value)
-  if param('certs', 'deploy').value && !File.file?(certs_tar.value)
+  if param('certs', 'deploy').value && !File.file?(certs_tar.value) || !Dir.exist?(ssl_directory.value)
```
ehelms: Would the ssl_directory be created by this point in all cases? I am thinking of a fresh install.

chris1984:
@ehelms is this good now?

hooks/pre_validations/12-check_capsule_tar.rb
```diff
@@ -4,12 +4,19 @@ def error(message)
...
+ssl_directory = param('certs', 'ssl_build_dir')
+certs_tar = param('foreman_proxy_certs', 'certs_tar') || param('foreman_proxy_content', 'certs_tar')
 
 if certs_tar.value
   certs_tar.value = File.expand_path(certs_tar.value)
-  if param('certs', 'deploy').value && !File.file?(certs_tar.value)
+end
+
+# On existing installs, check to see if ssl_directory exists and if so skip the certs_tar value check
+if File.file?('./config/last_scenario.yaml')
```
ekohl: This will fail in production. You may be able to use `Dir.join(scenario[:installer_dir], 'config', 'last_scenario.yaml')` but I'm wondering if this is really a good check.

ekohl:
I've taken another approach in https://github.com/Katello/katello-installer/pull/717 that I think is less fragile. It removes the certs answer after a successful run so on the next run it won't attempt to extract the certificate. Untested though.

chris1984:
Closing this one based on @ekohl comment and pr.

PR #718, opened 2018-11-27
Fixes #25564 - Generates proper output

patilsuraj767:
This issue was caused because the fix in the following  RFE.

[RFE] katello-certs-check to distinguish between Satellite and Capsule
https://projects.theforeman.org/issues/22694

Code snippet

```
HOSTNAME=$(hostname -f)
CERT_HOSTNAME=$(openssl x509 -noout -subject -in $CERT_FILE | sed -e 's/^subject.*CN=\([a-zA-Z0-9\.\-]*\).*$/\1/')

if [ $EXIT_CODE == "0" -a $CERT_HOSTNAME == $HOSTNAME ]; then
echo -e "${GREEN}Validation succeeded${RESET}\n"
    cat <<EOF

To install the Katello main server with the custom certificates, run:

    satellite-installer --scenario satellite\\
                      --certs-server-cert "$(readlink -f $CERT_FILE)"\\
                      --certs-server-key "$(readlink -f $KEY_FILE)"\\
                      --certs-server-ca-cert "$(readlink -f $CA_BUNDLE_FILE)"

```

Issue is we are checking the hostname against the cert CN and in case of wildcard cert CN is *.example.com. So this condition fails in case of wildcard certs that are to be installed in the satellite server. 

Need to modify the logic, better to give the user the ability to define target for which cert to be generated.

```
katello-certs-check -t foreman -c /customcerts/cert_crt.pem -k /customcerts/cert_key.pem -b /customcerts/CA_crt.pem
```


bin/katello-certs-check
```diff
@@ -11,12 +11,15 @@ function usage () {
 Verifies, that custom SSL certificate files are usable
 as part of the Katello installation. When passing filenames use absolute paths.
 
-usage: $0 -c CERT_FILE -k KEY_FILE -b CA_BUNDLE_FILE
+usage: $0 t- TARGET -c CERT_FILE -k KEY_FILE -b CA_BUNDLE_FILE
```
ekohl: Should this be `-t`?

bin/katello-certs-check
```diff
@@ -45,8 +48,11 @@ if [ -z "$CERT_FILE" -o -z "$KEY_FILE" -o -z "$CA_BUNDLE_FILE" ]; then
     exit 1
 fi
 
-HOSTNAME=$(hostname -f)
-CERT_HOSTNAME=$(openssl x509 -noout -subject -in $CERT_FILE | sed -e 's/^subject.*CN=\([a-zA-Z0-9\.\-]*\).*$/\1/')
+if [[ $TARGET != "foreman" && $TARGET != "foreman-proxy" ]]; then
+    echo 'Target not specified or incorrect' >&2
```
ekohl: The target usage doesn't tell the user that only `foreman` and `foreman-proxy` are the only valid values. The user now has no way to know.

bin/katello-certs-check
patilsuraj767: Yes

ehelms:
This will need a rebase please

PR #723, opened 2018-11-30
Fixes #25564 - Generates proper output

patilsuraj767:
This issue was caused because the fix in the following RFE.

[RFE] katello-certs-check to distinguish between Satellite and Capsule
https://projects.theforeman.org/issues/22694

Code snippet

```
HOSTNAME=$(hostname -f)
CERT_HOSTNAME=$(openssl x509 -noout -subject -in $CERT_FILE | sed -e 's/^subject.*CN=\([a-zA-Z0-9\.\-]*\).*$/\1/')

if [ $EXIT_CODE == "0" -a $CERT_HOSTNAME == $HOSTNAME ]; then
echo -e "${GREEN}Validation succeeded${RESET}\n"
    cat <<EOF

To install the Katello main server with the custom certificates, run:

    satellite-installer --scenario satellite\\
                      --certs-server-cert "$(readlink -f $CERT_FILE)"\\
                      --certs-server-key "$(readlink -f $KEY_FILE)"\\
                      --certs-server-ca-cert "$(readlink -f $CA_BUNDLE_FILE)"

```
Issue is we are checking the hostname against the cert CN and in case of wildcard cert CN is *.example.com. So this condition fails in case of wildcard certs that are to be installed in the satellite server.

Need to modify the logic, better to give the user the ability to define target for which cert to be generated.
```
katello-certs-check -t foreman -c /customcerts/cert_crt.pem -k /customcerts/cert_key.pem -b /customcerts/CA_crt.pem
```

### DIFF
--- a/bin/katello-certs-check
+++ b/bin/katello-certs-check
@@ -55,8 +55,6 @@ if [[ $TARGET != "foreman" && $TARGET != "foreman-proxy" ]]; then
     exit 1
 fi
 
-HOSTNAME=$(hostname -f)
-CERT_HOSTNAME=$(openssl x509 -noout -subject -in $CERT_FILE | sed -e 's/^subject.*CN=\([a-zA-Z0-9\.\-]*\).*$/\1/')
 
 function error () {
     echo -e "\n${RED}[FAIL]${RESET}\n"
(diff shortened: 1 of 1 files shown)


chris1984:
@ekohl want me to test this with the custom certs I have?

ekohl:

PR #727, opened 2018-12-10
Drop Foreman from Apache group on upgrade

ehelms:
This removes a dependency cycle caused by prior installations.

Reflects https://github.com/theforeman/puppet-katello/commit/4a0f1b22077d44d2cd2e5c068fa7d9750c0ee602

### DIFF
--- a/hooks/pre/30-upgrade.rb
+++ b/hooks/pre/30-upgrade.rb
@@ -163,6 +163,10 @@ def upgrade_qpid_paths
   end
 end
 
+def drop_apache_foreman_group
+  Kafo::Helpers.execute("gpasswd -d apache foreman")
+end
+
 def upgrade_step(step, options = {})
   noop = app_value(:noop) ? ' (noop)' : ''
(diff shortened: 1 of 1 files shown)


hooks/pre/30-upgrade.rb
```diff
@@ -228,5 +232,7 @@ def fail_and_exit(message)
     upgrade_step :remove_legacy_mongo
   end
 
+  upgrade_step :drop_apache_foreman_group
```
ekohl: This should be in the `if katello` block since it was in `puppet-katello`.

hooks/pre/15-check-disk.rb
```diff
@@ -13,7 +13,7 @@
       $stderr.puts DISK_SIZE
       kafo.class.exit 1
     end
-  rescue
+  rescue Exception => e
```
ekohl: https://github.com/Katello/katello-installer/pull/728

PR #729, opened 2018-12-11
Fixes #25665 - Use the bash comparisons can function calls

ekohl:
The use of == is a bash-only feature. Since we're using this already, we might as well go all the way and use the safer [[ ]] checks.

Using $() instead of `` is also a better practice because it allows nesting and slightly easier to read.

### DIFF
--- a/bin/katello-certs-check
+++ b/bin/katello-certs-check
@@ -66,11 +66,11 @@ function warning () {
 function check-server-cert-encoding () {
     printf 'Checking server certificate encoding: '
     openssl x509 -inform pem -in $CERT_FILE -noout -text &> /dev/null
-    if [ $? == "0" ]; then
+    if [[ $? == "0" ]]; then
         success
     else
         openssl x509 -inform der -in $CERT_FILE -noout -text &> /dev/null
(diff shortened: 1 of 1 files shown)


chris1984:
@ekohl i am getting this when running the script:

```
[root@satellite1 ~]# ./katello-certs-check -b ca.pem -k server.key -c server.crt 
Checking server certificate encoding: 
[OK]

Checking expiration of certificate: 
[OK]

Checking expiration of CA bundle: 
[OK]

Checking if server certificate has CA:TRUE flag 
[OK]

Checking to see if the private key matches the certificate: 
[OK]

Checking CA bundle against the certificate file: 
[OK]

Checking Subject Alt Name on certificate 
[OK]

Checking Key Usage extension on certificate for Key Encipherment 
[OK]

./katello-certs-check: line 168: syntax error in conditional expression
./katello-certs-check: line 168: syntax error near `-a'
./katello-certs-check: line 168: `if [[ $EXIT_CODE == "0" -a $CERT_HOSTNAME == $HOSTNAME ]]; then'
```

ekohl:
Thanks, will pick that up.

PR #733, opened 2019-07-29
[CP] Fixes #27424 - Ensure DB is seeded

jturel:
"cherry-pick" of https://github.com/theforeman/foreman-installer/pull/375

### DIFF
--- a/hooks/post/30-upgrade.rb
+++ b/hooks/post/30-upgrade.rb
@@ -1,5 +1,10 @@
 require 'fileutils'
 
+def db_seed
+  status = Kafo::Helpers.execute('foreman-rake db:seed')
+  fail_and_exit "Upgrade failed while seeding the DB" unless status
+end
+
 def upgrade_tasks
   status = Kafo::Helpers.execute('foreman-rake upgrade:run')
   fail_and_exit "Application Upgrade Failed" unless status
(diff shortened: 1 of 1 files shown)


jturel:
@ekohl here it is for 3.10

ekohl (approved):

hooks/post/30-upgrade.rb
```diff
@@ -12,7 +17,11 @@ def fail_and_exit(message)
 
 if app_value(:upgrade)
   if [0, 2].include?(@kafo.exit_code)
-    upgrade_tasks if Kafo::Helpers.module_enabled?(@kafo, 'foreman')
+    if module_enabled?('foreman')
```
evgeni, line 20: Does that work in 3.10 already?

ekohl:
I was wondering: do we still support 3.10? Foreman 1.20 has ended and we likely won't do another release. I was actually thinking about archiving this repository.

hooks/post/30-upgrade.rb
ekohl: Good point. It doesn't work as it should since it errors out if the module doesn't exist (like in the foreman-proxy-content scenario).

ekohl (changes requested):
Revoking approval based on what @evgeni said.

